Issue #31, opened 2019-06-27
Mechanism for adding multiple foreign key constraints at once


simonw:
Needed by [db-to-sqlite](https://github.com/simonw/db-to-sqlite). It currently works by collecting all of the foreign key relationships it can find and then applying them at the end of the process.

The problem is, the `add_foreign_key()` method looks like this:

https://github.com/simonw/sqlite-utils/blob/86bd2bba689e25f09551d611ccfbee1e069e5b66/sqlite_utils/db.py#L498-L516

That means it's doing a full `VACUUM` for every single relationship it sets up - and if you have hundreds of foreign key relationships in your database this can take hours.

I think the right solution is to have a `.add_foreign_keys(list_of_args)` method which does the bulk operation and then a single `VACUUM`. `.add_foreign_key(...)` can then call the bulk action with a single list item.

PR #32, opened 2019-06-28
db.add_foreign_keys() method

simonw:
Refs #31. Still TODO:

- [x] Unit tests
- [x] Documentation

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -289,7 +289,7 @@ class Database:
 
         # Construct SQL for use with "UPDATE sqlite_master SET sql = ? WHERE name = ?"
         table_sql = {}
-        for table, column, other_table, other_column in foreign_keys:
+        for table, column, other_table, other_column in foreign_keys_to_create:
             old_sql = table_sql.get(table, self[table].schema)
             extra_sql = ",\n   FOREIGN KEY({column}) REFERENCES {other_table}({other_column})\n".format(
                 column=column, other_table=other_table, other_column=other_column
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -252,6 +252,69 @@ class Database:
             )
         )
 
+    def add_foreign_keys(self, foreign_keys):
+        # foreign_keys is a list of explicit 4-tuples
+        assert all(
+            len(fk) == 4 and isinstance(fk, (list, tuple)) for fk in foreign_keys
+        ), "foreign_keys must be a list of 4-tuples, (table, column, other_table, other_column)"
+
+        foreign_keys_to_create = []
+
+        # Verify that all tables and columns exist
+        for table, column, other_table, other_column in foreign_keys:
+            if not self[table].exists:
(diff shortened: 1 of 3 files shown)


simonw:
Documentation, to be added as a subsection of https://sqlite-utils.readthedocs.io/en/latest/python-api.html#adding-foreign-key-constraints

### Adding multiple foreign key constraints at once

The final step in adding a new foreign key to a SQLite database is to run VACUUM, to ensure the new foreign key is available in future introspection queries.

VACUUM against a large (multi-GB) database can take several minutes or longer. If you are adding multiple foreign keys using table.add_foreign_key(...) these can quickly add up.

Instead, you can use db.add_foreign_keys(...) to add multiple foreign keys within a single transaction. This method takes a list of four-tuples, each one specifying a table, column, other_table and other_column.

Here's an example adding two foreign keys at once:

    db.add_foreign_keys([
        ("dogs", "breed_id", "breeds", "id"),
        ("dogs", "home_town_id", "towns", "id")
    ])

sqlite_utils/db.py
```diff
@@ -252,6 +252,69 @@ def create_view(self, name, sql):
...
+        # Construct list of args to be used with "UPDATE sqlite_master SET sql = ? WHERE name = ?"
+        sql_args = []
+        for table, column, other_table, other_column in foreign_keys:
+            old_sql = self[table].schema
+            extra_sql = ",\n   FOREIGN KEY({column}) REFERENCES {other_table}({other_column})\n".format(
+                column=column, other_table=other_table, other_column=other_column
+            )
+            # Stick that bit in at the very end just before the closing ')'
+            last_paren = old_sql.rindex(")")
+            new_sql = old_sql[:last_paren].strip() + extra_sql + old_sql[last_paren:]
```
simonw: This is wrong - adding multiple new foreign keys to the same table won't wrk as we need to accumulate the SQL.

sqlite_utils/db.py
```diff
@@ -252,6 +252,69 @@ def create_view(self, name, sql):
...
+                and fk.other_table == other_table
+                and fk.other_column == other_column
+            ):
+                foreign_keys_to_create.append(
+                    (table, column, other_table, other_column)
+                )
+
+        # Construct list of args to be used with "UPDATE sqlite_master SET sql = ? WHERE name = ?"
+        sql_args = []
+        for table, column, other_table, other_column in foreign_keys:
```
simonw: Should loop through foreign_keys_to_create here

sqlite_utils/db.py
simonw: Fixed in https://github.com/simonw/sqlite-utils/pull/32/commits/889e2afccf8a252915b9273cd0421bc880ac483b

sqlite_utils/db.py
simonw: Fixed in https://github.com/simonw/sqlite-utils/pull/32/commits/fc81588cc31df58374c996884c67cedd98a06f4a

Issue #35, opened 2019-06-30
table.update(...) method


simonw:
Spun off from #23 - this method will allow a user to update a specific row.

Currently the only way to do that it is to call `.upsert({full record})` with the primary key field matching an existing record - but this does not support partial updates.
```python
db["events"].update(3, {"name": "Renamed"})
```
This method only works on an existing table, so there's no need for a `pk="id"` specifier - it can detect the primary key by looking at the table.

If the primary key is compound the first argument can be a tuple:
```python
db["events_venues"].update((3, 2), {"custom_label": "Label"})
```
The method can be called without the second dictionary argument. Doing this selects the row specified by the primary key (throwing an error if it does not exist) and remembers it so that chained operations can be carried out - see proposal in https://github.com/simonw/sqlite-utils/issues/23#issuecomment-507055345

simonw:
The last touched primary key should go in `last_pk`.

https://github.com/simonw/sqlite-utils/blob/ba6bd47f15c154e348c3c85c05858053a4ab7bab/docs/python-api.rst#L126

simonw:
Should this support both dictionaries AND keyword arguments?
```python
db["events_venues"].update((3, 2), custom_label="Label")
```
Pros: this is quite nice to look at.
Cons: would have to continue supporting dictionaries as well, since otherwise you won't be able to update columns with names that are not valid Python identifiers. So now there are two ways.

I'm going to stick with dictionaries only for the moment.

Issue #36, opened 2019-07-01
Support compound primary keys


simonw:
This should work:
```python
table = db["dog_breeds"].insert({
    "dog_id": 1,
    "breed_id": 2
}, pk=("dog_id", "breed_id"))
```
Needed for m2m work in #23

PR #38, opened 2019-07-14
table.update() method

simonw:
Refs #35

Still to do:

- [x] Unit tests
- [x] Switch to using `.get()`
- [x] Better exceptions, plus unit tests for what happens if pk does not exist
- [x] Documentation
- [x] Ensure compound primary keys work properly
- [x] `alter=True` support


simonw:
I think I need to solve #36 (compound primary key support) first.

simonw:
This should switch to using `.get()`, introduced in https://github.com/simonw/sqlite-utils/commit/c65b67ca46f70e2da46a5b945f4ed358173262e9

Issue #39, opened 2019-07-14
table.get(...) method


simonw:
Utility method for fetching a record by its primary key.

Accepts a single value (for primary key / rowid tables) or a list/tuple of values (for compound primary keys, refs #36).

Raises a `NotFoundError` if the record cannot be found.

PR #40, opened 2019-07-15
.get() method plus support for compound primary keys

simonw:
- [x] Tests for the `NotFoundError` exception
- [x] Documentation for `.get()` method
- [x] Support `--pk` multiple times to define CLI compound primary keys
- [x] Documentation for compound primary keys


simonw:
This will close #36 and #39 once it is finished.

Issue #54, opened 2019-08-15
Ability to list views, and to access db["view_name"].rows / rows_where / etc


ftrain:
The docs show me how to create a view via `db.create_view()` but I can't seem to get back to that view post-creation; if I query it as a table it returns `None`, and it doesn't appear in the table listing, even though querying the view works fine from inside the sqlite3 command-line.

It'd be great to have the view as a pseudo-table, or if the python/sqlite3 module makes that hard to pull off (I couldn't figure it out), to have that edge-case documented next to the `db.create_view()` docs.

simonw:
Good catch: `create_view()` is an earlier utility method from before I started adding methods like `db["table"].rows` and `.rows_where()`.

simonw:
Closed by #55

simonw:
Documentation is here: https://sqlite-utils.readthedocs.io/en/latest/python-api.html#listing-views

simonw:
Released in 1.10: https://sqlite-utils.readthedocs.io/en/stable/changelog.html#v1-10

ftrain:
Extremely cool and easy to understand. Thank you!

PR #55, opened 2019-08-17
Ability to introspect and run queries against views

simonw:
See #54


simonw:
I still need to implement the part of this where certain actions against views (like updating/inserting) are not allowed.

I think the better way to do that is to introduce a new class - `View` - rather than reusing the existing `Table` class - since there's a lot of stuff on Table (like primary key / foreign key introspection) which simply doesn't make sense for views.

This will get a bit more complicated - maybe I need both Table and View to extend a common subclass here?

PR #57, opened 2019-09-02
Add triggers while enabling FTS

amjith:
This adds the option for a user to set up triggers in the database to keep their FTS table in sync with the parent table. 

Ref: https://sqlite.org/fts5.html#external_content_and_contentless_tables

I would prefer to make the creation of triggers the default behavior, but that will break existing usage where people have been calling `populate_fts` after inserting new rows.

I am happy to make changes to the PR as you see fit.

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -266,12 +266,12 @@ def create_index(path, table, column, name, unique, if_not_exists):
 @click.option("--fts4", help="Use FTS4", default=False, is_flag=True)
 @click.option("--fts5", help="Use FTS5", default=False, is_flag=True)
 @click.option(
-    "--create-update-triggers",
+    "--create-triggers",
     help="Create triggers to update the FTS tables when the parent table changes.",
     default=False,
     is_flag=True,
 )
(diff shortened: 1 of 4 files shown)


simonw:
This is brilliant!

One tiny suggestion: I like `--create-triggers` and `create_triggers=True` better for this as they are shorter but still capture what it dies - especially since the underlying SQL uses `CREATE TRIGGER` statements.

amjith:
I have merged the other PR (#56) into this one. 

I have incorporated your suggestions. Cheers!

simonw:
This is fantastic, thanks so much.

I spotted a Datasette bug triggered by this: Datasette has its own version of the `detect_fts` function - at https://github.com/simonw/datasette/blob/d224ee2c98ac39c2c6e21a0ac0c62e5c3e1ccd11/datasette/utils/__init__.py#L466-L479 - which fails to pick up FTS tables created using the new escaping pattern.

It's a bug in Datasette, not sqlite-utils - so I'll open an issue there.

simonw:
Released in 1.11 - thanks again! https://sqlite-utils.readthedocs.io/en/latest/changelog.html#v1-11

PR #75, opened 2020-01-07
Explicitly include tests and docs in sdist

jayvdb:
Also exclude 'tests' from runtime installation.


simonw:
This makes sense, thanks!

PR #80, opened 2020-01-31
on_create mechanism for after table creation

simonw:
I need this for `geojson-to-sqlite`, in particular https://github.com/simonw/geojson-to-sqlite/issues/6

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -219,6 +219,7 @@ class Database:
         defaults=None,
         hash_id=None,
         extracts=None,
+        on_create=None,
     ):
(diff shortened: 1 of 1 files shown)


simonw:
Still needs documentation and tests. Also I'm not certain that this should be an argument you can pass to the `.table()` constructor, need to think that over.

simonw:
Perhaps this should be called `after_create` instead of `on_create`.

simonw:
I think it does make sense to be able to pass it to the `.table()` constructor.

simonw:
This may not be the right feature after all, see https://github.com/simonw/geojson-to-sqlite/issues/6#issuecomment-580569002

simonw:
Ditching this since it won't actually solve my problem.

PR #96, opened 2020-03-29
Add type conversion for Panda's Timestamp

b0b5h4rp13:
Add type conversion for Panda's Timestamp, if Panda library is present in system
(thanks for this project, I was about to do the same thing from scratch)


simonw:
Test failure was just a Black formatting issue.

simonw:
I wonder if there are any other Pandas conversions we should be doing?

PR #120, opened 2020-07-08
Fix query command's support for DML

tsibley:
See commit messages for details. I ran into this while investigating another feature/issue.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1126,3 +1126,6 @@ def test_query_update(db_path, args, expected):
         cli.cli, [db_path, "update dogs set age = 5 where name = 'Cleo'"] + args
     )
     assert expected == result.output.strip()
+    assert db.execute_returning_dicts("select * from dogs") == [
+        {"id": 1, "age": 5, "name": "Cleo"},
+    ]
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -1126,3 +1126,6 @@ def test_query_update(db_path, args, expected):
         cli.cli, [db_path, "update dogs set age = 5 where name = 'Cleo'"] + args
     )
     assert expected == result.output.strip()
+    assert db.execute_returning_dicts("select * from dogs") == [
+        {"id": 1, "age": 5, "name": "Cleo"},
+    ]
```
tsibley, line 1131: In other tests, the order of the expected vs. gotten values in the asserts is reversed. However, pytest expects that the LHS is the gotten value and the RHS is the expected value and its error/debugging messages reflect that.

simonw: I had no idea! I've been habitually doing it the other way round because `1 = a` is a syntax error whereas `a = 1` is not (a typo for `a == 1`).

It sounds like I'm going to have to break that habit and switch to `actual == expected`.  Thanks for letting me know!

simonw:
This is an excellent fix, thanks!

PR #156, opened 2020-09-09
Typos in tests

simonwiles:
One of these is my fault, and the other is one I just happened to come across.  They're harmless, but might as well be fixed.


simonw:
Good spot, thanks.

Issue #114, opened 2020-05-20
table.transform() method for advanced alter table


simonw:
SQLite's `ALTER TABLE` can only do the following:

* Rename a table
* Rename a column
* Add a column

Notably, it cannot drop columns - so tricks like "add a float version of this text column, populate it, then drop the old one and rename" won't work.

The docs here https://www.sqlite.org/lang_altertable.html describe a way of implementing full alters safely within a transaction, but it's fiddly.

1. Create new table
2. Copy data
3. Drop old table
4. Rename new into old 

It would be great if `sqlite-utils` provided an abstraction to help make these kinds of changes safely.

simonw:
Idea: use a chained API to define a complex transition and then execute it all at once. For example:
```python
db["mytable"].transform().rename("col1", "col_1") \
  .change_type("col1", float) \
  .execute()
```

simonw:
Ideally this would all happen in a single transaction, such that other processes talking to the database would not see any inconsistent state while the table copy was taking place. Need to confirm that this is possible. Also refs transactions thoughts in #121.

simonw:
I'm not so keen on that chained API - it's pretty complicated.

Here's an idea for a much simpler interface. Essentially it lets you say "take table X and migrate its contents to a new table with this structure - then atomically rename the tables to switch them":
```python
db["mytable"].migrate_table({"id": int, "name": str"}, pk="id")
```
The `migrate_table()` method would take the same exact signature as the `table.create()` method: https://github.com/simonw/sqlite-utils/blob/a236a6bc771a5a6a9d7e814f1986d461afc422d2/sqlite_utils/db.py#L615-L625

simonw:
Alternative possible names:
- `.transform_table()`
- `.migrate()`
- `.transform()`

I'm torn between `.migrate_table()` and `.transform_table()`.

simonw:
Since neither the term "transform" or "migrate" are used in the codebase at the moment, I think I'll go with `.transform_table()` - that leaves the term "migrate" available for any future database migrations system (similar to Django's).

simonw:
Don't forget this step:
>  If foreign key constraints are enabled, disable them using PRAGMA foreign_keys=OFF.

simonw:
Thinking about the method signature:
```python
    def transform_table(
        self,
        columns,
        pk=None,
        foreign_keys=None,
        column_order=None,
        not_null=None,
        defaults=None,
        hash_id=None,
        extracts=None,
[...]

PR #161, opened 2020-09-21
table.transform() method

simonw:
Refs #114

- [x] Ability to change the primary key
- [x] Support for changing default value for columns
- [x] Support for changing `NOT NULL` status of columns
- [x] Support for copying existing foreign keys and removing them
- <strike>Support for `conversions=` parameter</strike>
- [x] Detailed documentation
- [x] `PRAGMA foreign_keys` stuff

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -1415,3 +1415,25 @@ You can use it in code like this:
     if spatialite:
         db.conn.enable_load_extension(True)
         db.conn.load_extension(spatialite)
+
+.. _python_api_register_function:
+
+Registering custom SQL functions
+================================
+
+SQLite supports registering custom SQL functions written in Python. The ``@db.register_function`` decorator provides syntactic sugar for registering these functions.
+
+.. code-block:: python
+
+    from sqlite_utils import Database
+
+    db = Database(memory=True)
+
+    @db.register_function
+    def reverse_string(s):
+        return "".join(reversed(list(s)))
+
+    print(db.execute('select reverse_string("hello")').fetchone())[0]
+    # This prints "olleh"
+
+If your Python function accepts multiple arguments the SQL version will accept the same number of arguments.
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -908,6 +908,99 @@ You can drop a table or view using the ``.drop()`` method:
 
     db["my_table"].drop()
 
+.. _python_api_transform:
+
+Transforming a table
+====================
+
+The SQLite ``ALTER TABLE`` statement is limited. It can add columns and rename tables, but it cannot rename columns, drop columns, change column types, change ``NOT NULL`` status or change the primary key for a table.
+
+The ``table.transform()`` method can do all of these things, by implementing a multi-step pattern `described in the SQLite documentation <https://www.sqlite.org/lang_altertable.html#otheralter>`__:
+
+1. Start a transaction
+2. ``CREATE TABLE tablename_new_x123`` with the required changes
+3. Copy the old data into the new table using ``INSERT INTO tablename_new_x123 SELECT * FROM tablename;``
+4. ``DROP TABLE tablename;``
+5. ``ALTER TABLE tablename_new_x123 RENAME TO tablename;``
+6. Commit the transaction
+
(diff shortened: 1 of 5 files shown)


simonw:
I still need to figure out what to do about these various other table properties: https://github.com/simonw/sqlite-utils/blob/b34c9b40c206d7a9d7ee57a8c1f198ff1f522735/sqlite_utils/db.py#L775-L787

simonw:
When you transform a table, it should keep its primary key, foreign keys, not_null and defaults. I don't think it needs to care about `hash_id` or `extracts=` since those don't affect the structure of the table as it is being created - well, `hash_id` does but if we are transforming an existing table we will get the `hash_id` column for free.

simonw:
How would I detect which columns are `not_null` and what their defaults are? I don`t think my introspection logic handles that yet.

simonw:
Good news: the `.columns` introspection does tell me those things:
```
>>> import sqlite_utils
>>> db.create_table("foo", {"id": int, "name": str, "age": int}, defaults={"age": 1}, not_null={"name", "age"})
<Table foo (id, name, age)>
>>> db["foo"]
<Table foo (id, name, age)>
>>> print(db["foo"].schema)
CREATE TABLE [foo] (
   [id] INTEGER,
   [name] TEXT NOT NULL,
   [age] INTEGER NOT NULL DEFAULT 1
)
>>> db["foo"].columns
[Column(cid=0, name='id', type='INTEGER', notnull=0, default_value=None, is_pk=0), Column(cid=1, name='name', type='TEXT', notnull=1, default_value=None, is_pk=0), Column(cid=2, name='age', type='INTEGER', notnull=1, default_value='1', is_pk=0)]
```

simonw:
I should support `not_null=` and `default=` arguments to the `.transform()` method because it looks like you can't use `ALTER TABLE` to change those.

simonw:
Syntax challenge: I could use `.transform(defaults={"age": None})` to indicate that the `age` column should have its default removed, but how would I tell `.transform()` that the `age` column, currently `not null`, should have the `not null` removed from it?

simonw:
A less horrible interface might be the following:
```python
# Set the 'age' column to be not_null:
table.transform(not_null={"age"})
# The 'age' column is not null but I don't want it to be:
table.transform(not_null={"age": False})
```
So if the argument is a set it means "make sure these are all not null" - if the argument is a dictionary it means "set these to be null or not null depending on if their dictionary value is true or false".

simonw:
Idea: allow a `conversions=` parameter, as seen on `.insert_all()` and friends, which lets you apply a SQL transformation function as part of the operation. E.g.:

```python
table.transform({"age": int}, conversions={"name": "upper(?)"})
```

simonw:
I'm not going to do `conversions=` because it would be inconsistent with how they work elsewhere. The SQL generated by this function looks like this:

    INSERT INTO dogs_new_tmp VALUES (a, b) SELECT a, b from dogs;

So passing `conversions={"name": "upper(?)"})` wouldn't make sense, since we're not using arguments hence there is no-where for that `?` to go.

simonw:
Design decision needed on foreign keys: what does the syntax look like for removing an existing foreign key?

Since I already have a good implementation of `add_foreign_key()` I'm tempted to only support dropping them. Maybe like this:

```python
table.transform(drop_foreign_keys=[("author_id", "author", "id")])
```
It's a bit crufty but it's such a rare use-case that I think this will be good enough.

simonw:
Just needs documentation now.

simonw:
It would be neat if `.transform(pk=None)` converted a primary key table to a rowid table.

simonw:
One last thing. https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_change says that the first step should be:

>  If foreign key constraints are enabled, disable them using PRAGMA foreign_keys=OFF. 

And the last steps should be:

> If foreign key constraints were originally enabled then run PRAGMA foreign_key_check to verify that the schema change did not break any foreign key constraints.
>
> Commit the transaction started in step 2.
>
> If foreign keys constraints were originally enabled, reenable them now. 

I need to implement that.

Issue #172, opened 2020-09-23
Improve performance of extract operations


simonw:
This command took about 12 minutes (against a 150MB file with 680,000 rows in it):
```
sqlite-utils extract salaries.db salaries \
   'Organization Group Code' 'Organization Group' \
  --table 'organization_groups' \
  --fk-column 'organization_group_id' \
  --rename 'Organization Group Code' code \
  --rename 'Organization Group' name
```
I'm pretty confident we can do better than that.

simonw:
Steps to produce that database:
```
curl -o salaries.csv 'https://data.sfgov.org/api/views/88g8-5mnd/rows.csv?accessType=DOWNLOAD'
sqlite-utils insert salaries.db salaries salaries.csv --csv
```

simonw:
Here's the loop that's taking the time: https://github.com/simonw/sqlite-utils/blob/1ebffe1dbeaed7311e5b61ed988f4cd701e84808/sqlite_utils/db.py#L892-L897

simonw:
`lookup_table.lookup(lookups)` is doing a SQL lookup. This could be cached in-memory, maybe with a LRU cache, to avoid looking up the primary key for records that we have recently used.

The `.update()` method it is calling first does a `get()` and then does a SQL `UPDATE ... WHERE`:

https://github.com/simonw/sqlite-utils/blob/1ebffe1dbeaed7311e5b61ed988f4cd701e84808/sqlite_utils/db.py#L1244-L1264

Batching those updates may have an effect. Or finding a way to skip the `.get()` since we already know we have a valid record.

simonw:
I ran `sudo py-spy top -p 123` against the process while it was running and the most time is definitely spent in `.update()`:
```
Total Samples 1000
GIL: 0.00%, Active: 90.00%, Threads: 1

  %Own   %Total  OwnTime  TotalTime  Function (filename:line)                                                                                                                                  
 38.00%  38.00%    3.85s     3.85s   update (sqlite_utils/db.py:1283)
 27.00%  27.00%    2.12s     2.12s   execute (sqlite_utils/db.py:161)
 10.00%  10.00%   0.890s    0.890s   execute (sqlite_utils/db.py:163)
 10.00%  17.00%   0.870s     1.54s   columns (sqlite_utils/db.py:553)
  0.00%   0.00%   0.110s    0.210s   <listcomp> (sqlite_utils/db.py:554)
[...]

PR #174, opened 2020-09-24
Much, much faster extract() implementation

simonw:
Takes my test down from ten minutes to four seconds. Refs #172.

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -901,15 +901,15 @@ class Table(Queryable):
             for col, typ in self.columns_dict.items()
             if col in columns
         }
-        if lookup_table.exists() and not set(lookup_columns_definition.keys()).issubset(
-            lookup_table.columns_dict.keys()
-        ):
-            # TODO: Write test for this
-            raise InvalidColumns(
-                "Lookup table {} already exists but does not have columns {}".format(
-                    table, lookup_columns_definition.keys()
+        if lookup_table.exists():
+            if not set(lookup_columns_definition.items()).issubset(
+                lookup_table.columns_dict.items()
+            ):
+                raise InvalidColumns(
+                    "Lookup table {} already exists but does not have columns {}".format(
+                        table, lookup_columns_definition
+                    )
                 )
-            )
         else:
             lookup_table.create(
                 {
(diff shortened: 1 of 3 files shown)


simonw:
This could do with a little bit more testing - I'm worried there may be column or table name edge cases that are not covered yet. I also need to remove the progress bar code since that no longer makes sense for this implementation.

simonw:
After running through the steps in https://simonwillison.net/2020/Sep/23/sqlite-utils-extract/ I get a table that looks like this:

salaries__salaries__683_558_rows

The foreign key columns are all at the end of the table. It would be nicer if they were arranged in the same order as the columns they replaced.

simonw:
`import functools` is no longer needed.

simonw:
I could also be a bit smarter about transaction handling. I think it may be possible to run this entire operation in a single transaction now.

simonw:
The way the lookup table works here differs from the previous implementation. In the previous implementation the usage of `.lookup()` meant that an existing table would be modified to fit the new purpose. That no longer happens in this version. Need to make a design decision about how this should work.

It should definitely be possible to use an existing lookup table - imagine a database where several tables have a "Departments" column and we want to extract all of those values out to a single shared "Departments" table.

simonw:
I may revert the now unnecessary undocumented tweaks to the `.update()` method made in 66d506587eba9f0715267d6560b97c1fa44cc781 as well.

simonw:
For reusing the lookup table: I'm going to raise an error if a lookup table exists but without the correct columns. The caller can then add those columns and try again.

sqlite_utils/db.py
```diff
@@ -886,38 +892,85 @@ def extract(self, columns, table=None, fk_column=None, rename=None, progress=Non
...
+        # Populate the lookup table with all of the extracted unique values
+        lookup_columns_definition = {
+            (rename.get(col) or col): typ
+            for col, typ in self.columns_dict.items()
+            if col in columns
+        }
+        if lookup_table.exists() and not set(lookup_columns_definition.keys()).issubset(
+            lookup_table.columns_dict.keys()
+        ):
+            # TODO: Write test for this
```
simonw, line 907: Still need to write this test.

sqlite_utils/db.py
```diff
@@ -731,6 +730,7 @@ def transform(
         not_null=None,
         defaults=None,
         drop_foreign_keys=None,
+        column_order=None,
```
simonw, line 733: Need to update docs for `.transform()` now that `column_order=` is available.

sqlite_utils/db.py
simonw: The tests should also cover the case of reusing a lookup table.

PR #178, opened 2020-09-25
Update README.md

shakeel:
The `sqlite-utils insert releases.db releases - --pk` is missing the pk field name, added ` "id"` to fix it.


simonw:
Thanks for the fix!

PR #189, opened 2020-10-26
Allow iterables other than Lists in m2m records

adamwolf:
I was playing around with sqlite-utils, creating a Roam Research dogsheep-style importer for Datasette, and ran into a slight snag.

I wanted to use a generator to add an order column in an importer. It looked something like:

```
def order_generator(iterable, attr=None):
    if attr is None:
        attr = "order"
    order: int = 0

    for i in iterable:
        i[attr] = order
        order += 1
        yield i
```

When I used this with `insert_all` and other things, it worked fine--but it didn't work as the `records` argument to `m2m`.  I dug into it, and sqlite-utils is explicitly checking if the records argument is a list or a tuple.  I flipped the check upside down, and now it checks if the argument is a mapping.  If it's a mapping, it wraps it in a list, otherwise it leaves it alone.

(I get that it might not really make sense to put the order column on the second table.  I changed my import schema a bit, and no longer have a real example, but maybe this change still makes sense.)

The automated tests still pass, but I did not add any new ones.

Let me know what you think! I'm really loving Datasette and its ecosystem; thanks for everything!

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -1,5 +1,6 @@
 from .utils import sqlite3, OperationalError, suggest_column_types, column_affinity
 from collections import namedtuple, OrderedDict
+from collections.abc import Mapping
 import contextlib
 import datetime
 import decimal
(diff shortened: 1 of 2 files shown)


simonw:
This is a great fix, thanks! If you add a unit test somewhere in here I'll merge the PR:  https://github.com/simonw/sqlite-utils/blob/main/tests/test_m2m.py

adamwolf:
No problem. I added a test. Let me know if it looks sufficient or if you want me to to tweak something!

If you don't mind, would you tag this PR as "hacktoberfest-accepted"? If you do mind, no problem and I'm sorry for asking :) My kiddos like the shirts.

simonw:
This is great, thank you very much.

Issue #192, opened 2020-11-03
sqlite-utils search command


simonw:
A command that knows how to run a search against a FTS enabled table and return results ranked by relevance.

simonw:
Relevant example using a SQLite CTE: https://github.com/simonw/datasette/issues/268#issuecomment-675725464

CTEs were added in SQLite 3.8.3 in 2014-02-03 so they should be safe to use. If someone tries to run `sqlite-utils search` no an older version of SQLite they'll get an error, which I think is OK.

simonw:
Just realized there's a problem with the SQL I am using here: joining to get `rank` in this way only works against FTS5 tables, it doesn't work against FTS4.

 https://github.com/simonw/sqlite-utils/blob/c8a900df59efd34f394c863c0adff9912f1bf1d7/sqlite_utils/db.py#L1303-L1319

simonw:
I could depend on my `sqlite-fts4` library to solve this: https://github.com/simonw/sqlite-fts4

Or I could say that only `FTS5` is supported for ranked searches - but still support `.search()` against FTS4 just without the option to sort by relevance.

simonw:
FTS5 was added in SQLite 3.9.0 in 2015-10-14 - so about a year after CTE, which means CTEs will always be safe to use with FTS5 queries.

simonw:
I want `.search()` to work against both FTS5 and FTS4 tables - but sort by rank should only work for FTS5.

This means I need to be able to introspect and tell if a table is FTS4 or FTS5.

simonw:
This is a bit surprising:
```
(sqlite-utils) sqlite-utils % sqlite-utils search 24ways-fts4.db articles maps -c title      
[{"rowid": 41, "title": "What Is Vagrant and Why Should I Care?", "rank": -1.9252039178908076},
 {"rowid": 298, "title": "First Steps in VR", "rank": -1.9945466378736434},
 {"rowid": 43, "title": "Content Production Planning", "rank": -2.1928058363046143},
 {"rowid": 100, "title": "Moo'y Christmas", "rank": -2.2698482999851675},
 {"rowid": 91, "title": "Infinite Canvas: Moving Beyond the Page", "rank": -2.290928999035195},
 {"rowid": 175, "title": "Front-End Code Reusability with CSS and JavaScript", "rank": -2.498731782924352},
 {"rowid": 209, "title": "Feeding the Audio Graph", "rank": -2.619968930100356},
[...]

PR #195, opened 2020-11-03
table.search() improvements plus sqlite-utils search command

simonw:
Refs #192. Still needs tests.

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -7,6 +7,7 @@ import hashlib
 import pathlib
 import sqlite_utils
 from sqlite_utils.db import AlterError
+import textwrap
 import itertools
 import json
 import os
@@ -45,7 +46,7 @@ def output_options(fn):
                 is_flag=True,
                 default=False,
             ),
-            click.option("-c", "--csv", is_flag=True, help="Output CSV"),
+            click.option("--csv", is_flag=True, help="Output CSV"),
             click.option("--no-headers", is_flag=True, help="Omit CSV headers"),
             click.option("-t", "--table", is_flag=True, help="Output as a table"),
             click.option(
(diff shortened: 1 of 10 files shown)


simonw:
Documentation so far: https://github.com/simonw/sqlite-utils/blob/6cadc6103ff1ba58c6409ce7fba74259e72965d9/docs/cli.rst#executing-searches

sqlite_utils/cli.py
```diff
@@ -968,6 +969,114 @@ def query(
...
+    columns = "*"
+    if column:
+        # Check they all exist
+        for c in column:
+            if c not in table_obj.columns_dict:
+                raise click.ClickException(
+                    "Table '{}' has no column '{}".format(dbtable, c)
+                )
+        columns = ", ".join("[{}]".format(c) for c in column)
+    sql = textwrap.dedent(
```
simonw, line 1031: I'd like this logic to live in `db.py`, not be exclusive to the CLI tool.

simonw:
I think I can match on `' ... ignore-two-single-quotes-next-to-each-other ... '` using a negative lookahead assertion `(?!...)`.

simonw:
Latest test failure:
```
114  ->	        assert [("racoons are biting trash pandas", "USA", "bar")] == table.search(
115  	            "bite", order="rowid"
116  	        )
117  	
118  	
119  	def test_optimize_fts(fresh_db):
(Pdb) table.search("bite")
[(2, 'racoons are biting trash pandas', 'USA', 'bar', -9.641434262948206e-07)]
```
The problem here is that the `table.search()` method now behaves differently for FTS4 v.s. FTS5 tables.

With FTS4 you get back just the table columns.

With FTS5 you also get back the `rowid` as the first column and the `rank` score as the last column.

This is weird. It also makes me question whether having `.search()` return a list of tuples is the right API design.

simonw:
Thanks to #198 (introducing a `rank_bm25()` custom function for FTS4) this feature will be able to offer relevance search for both FTS5 AND FTS4 tables.

Issue #117, opened 2020-06-23
Support for compound (composite) foreign keys


simonw:
It turns out SQLite supports composite foreign keys: https://www.sqlite.org/foreignkeys.html#fk_composite

Their example looks like this:
```sql
CREATE TABLE album(
  albumartist TEXT,
  albumname TEXT,
  albumcover BINARY,
  PRIMARY KEY(albumartist, albumname)
);

CREATE TABLE song(
  songid     INTEGER,
  songartist TEXT,
  songalbum TEXT,
  songname   TEXT,
  FOREIGN KEY(songartist, songalbum) REFERENCES album(albumartist, albumname)
);
```

Here's what that looks like in sqlite-utils:

```
In [1]: import sqlite_utils                                                                                                                

In [2]: import sqlite3                                                                                                                     

In [3]: conn = sqlite3.connect(":memory:")                                                                                                 

In [4]: conn                                                                                                                               
Out[4]: <sqlite3.Connection at 0x1087186c0>

In [5]: conn.executescript(""" 
   ...: CREATE TABLE album( 
   ...:   albumartist TEXT, 
   ...:   albumname TEXT, 
   ...:   albumcover BINARY, 
   ...:   PRIMARY KEY(albumartist, albumname) 
   ...: ); 
   ...:  
   ...: CREATE TABLE song( 
   ...:   songid     INTEGER, 
   ...:   songartist TEXT, 
   ...:   songalbum TEXT, 
   ...:   songname   TEXT, 
   ...:   FOREIGN KEY(songartist, songalbum) REFERENCES album(albumartist, albumname) 
   ...: ); 
   ...: """)                                                                                                                               
Out[5]: <sqlite3.Cursor at 0x1088def10>

In [6]: db = sqlite_utils.Database(conn)                                                                                                   

In [7]: db.tables                                                                                                                          
Out[7]: 
[<Table album (albumartist, albumname, albumcover)>,
 <Table song (songid, songartist, songalbum, songname)>]

In [8]: db.tables[0].foreign_keys                                                                                                          
Out[8]: []

In [9]: db.tables[1].foreign_keys                                                                                                          
Out[9]: 
[ForeignKey(table='song', column='songartist', other_table='album', other_column='albumartist'),
 ForeignKey(table='song', column='songalbum', other_table='album', other_column='albumname')]
```
The table appears to have two separate foreign keys, when actually it has a single compound composite foreign key.

simonw:
Here's what's missing:
```
[...]

PR #203, opened 2020-11-16
changes to allow for compound foreign keys

drkane:
Add support for compound foreign keys, as per issue #117 

Not sure if this is the right approach. In particular I'm unsure about:

 - the new `ForeignKey` class, which replaces the namedtuple in order to ensure that `column` and `other_column` are forced into tuples. The class does the job, but doesn't feel very elegant.
 - I haven't rewritten `guess_foreign_table` to take account of multiple columns, so it just checks for the first column in the foreign key definition. This isn't ideal.
 - I haven't added any ability to the CLI to add compound foreign keys, it's only in the python API at the moment.

The PR also contains a minor related change that columns and tables are always quoted in foreign key definitions.

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -69,7 +69,9 @@ class ForeignKey(ForeignKeyBase):
             other_column = tuple(other_column)
         else:
             other_column = (other_column,)
-        self = super(ForeignKey, cls).__new__(cls, table, column, other_table, other_column)
+        self = super(ForeignKey, cls).__new__(
+            cls, table, column, other_table, other_column
+        )
         return self
 
     @property
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -67,31 +67,37 @@ class ForeignKey:
             self.other_column = (other_column,)
         elif isinstance(other_column, (tuple, list)):
             self.other_column = tuple(other_column)
-    
+
     @property
     def column_str(self):
         return ",".join(["[{}]".format(c) for c in self.column])
-    
+
     @property
     def other_column_str(self):
         return ",".join(["[{}]".format(c) for c in self.other_column])
 
     def __eq__(self, other):
         if isinstance(other, ForeignKey):
-            return all((
-                self.table == other.table,
-                self.column == other.column,
-                self.other_table == other.other_table,
-                self.other_column == other.other_column,
-            ))
+            return all(
+                (
+                    self.table == other.table,
+                    self.column == other.column,
+                    self.other_table == other.other_table,
+                    self.other_column == other.other_column,
+                )
+            )
         return False
 
     def __lt__(self, other):
         if isinstance(other, ForeignKey):
-            return ((self.table, self.column, self.other_table, self.other_column) < 
-                    (other.table, other.column, other.other_table, other.other_column))
+            return (self.table, self.column, self.other_table, self.other_column) < (
+                other.table,
+                other.column,
+                other.other_table,
+                other.other_column,
+            )
         return False
-    
+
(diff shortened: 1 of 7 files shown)


simonw:
Sorry for not reviewing this yet! I'll try to carve out time to look at it in the next few days.

sqlite_utils/db.py
```diff
@@ -64,13 +64,46 @@
...
     ),
 )
-ForeignKey = namedtuple(
-    "ForeignKey", ("table", "column", "other_table", "other_column")
-)
 Index = namedtuple("Index", ("seq", "name", "unique", "origin", "partial", "columns"))
 Trigger = namedtuple("Trigger", ("name", "table", "sql"))
 
 
+class ForeignKey(ForeignKeyBase):
```
simonw, line 71: It looks like `ForeignKeyBase` is missing - I added this and the tests started passing again:
```python
ForeignKeyBase = namedtuple(
    "ForeignKey", ("table", "column", "other_table", "other_column")
)
```

simonw:
Sorry for taking so long to review this!

This approach looks great to me - being able to optionally pass a tuple anywhere the API currently expects a column is smart, and it's consistent with how the `pk=` parameter works elsewhere.

There's just one problem I can see with this: the way it changes the `ForeignKey(...)` interface to always return a tuple for `.column` and `.other_column`, even if that tuple only contains a single item.

This represents a breaking change to the existing API - any code that expects `ForeignKey.column` to be a single string (which is any code that has been written against that) will break.

As such, I'd have to bump the major version of `sqlite-utils` to `4.0` in order to ship this.

Ideally I'd like to make this change in a way that doesn't represent an API compatibility break. I need to think a bit harder about how that might be achieved.

simonw:
One way that this could avoid a breaking change would be to have `fk.column` and `fk.other_column` remain as strings for non-compound-foreign-keys, but turn into tuples for a compound foreign key.

This is a bit of an ugly API design, and it could still break existing code that encounters a compound foreign key for the first time - but it would leave code working for the more common case of a non-compound-foreign-key.

simonw:
Another option: expand the `ForeignKey` object to have `.columns` and `.other_columns` properties in addition to the existing `.column` and `.other_column` properties. These new plural properties would always return a tuple, which would be a one-item tuple for a non-compound-foreign-key.

The question then is what should `.column` and `.other_column` return for compound foreign keys?

I'd be inclined to say they should return `None` - which would trigger errors in code that encounters a compound foreign key for the first time, but those errors would at least be a strong indicator as to what had gone wrong.

We can label `.column` and `.other_column` as deprecated and then remove them in `sqlite-utils 4.0`.

Since this would still be a breaking change in some minor edge-cases I'm thinking maybe 4.0 needs to happen in order to land this feature. I'm not opposed to doing that, I was just hoping it might be avoidable.

drkane:
Thanks for looking at this - home schooling kids has prevented me from replying. 

I'd struggled with how to adapt the API for the foreign keys too - I definitely tried the String/Tuple approach. I hadn't considered the breaking changes that would introduce though. I can take a look at this and try and make the change - see which of your options works best.

I've got a workaround for the use-case I was looking at this for, so it wouldn't be a problem for me if it was put on the back burner until a hypothetical v4.0 anyway.

psychemedia:
Is there any progress elsewhere on the handling of compound / composite foreign keys, or is this PR still effectively open?

fgregg:
i'll adopt this PR to make the changes @simonw suggested https://github.com/simonw/sqlite-utils/pull/203#issuecomment-753567932

simonw:
Superseded by #770 and shipped in 4.0.

PR #204, opened 2020-11-29
use jsonify_if_need for sql updates

mfa:


simonw:
Great catch, thank you.

PR #224, opened 2021-01-22
Add fts offset docs.

polyrand:
The limit can be passed as a string to the query builder to have an offset. I have tested it using the shorthand `limit=f"15, 30"`, the standard syntax should work too.


polyrand:
I'm using the FTS methods in sqlite-utils for this website: [excalidraw.art](https://excalidraw.art/). I wanted to get pagination to have some kind of infinite scrolling in the landing page, and I ended up using that.

simonw:
I'm going to add a `offset=` parameter to support this case. Thanks for the suggestion!

PR #254, opened 2021-04-13
Fix incorrect create-table cli description

robjwells:
The description for `create-table` was duplicated from `create-index`.


simonw:
Thanks!

Issue #257, opened 2021-04-26
Insert from JSON containing strings with non-ascii characters are escaped as unicode for lists, tuples, dicts.


dylan-wu:
JSON Test File (test.json):

```json
[
    {
        "id": 123,
        "text": "FR Théâtre"
    },
    {
        "id": 223,
        "text": [
            "FR Théâtre"
        ]
    }
]
```

Command to import:

```bash
sqlite-utils insert test.db text test.json --pk=id
```

Resulting table view from datasette:

![image](https://user-images.githubusercontent.com/6586811/116147833-cdf2fb00-a6a5-11eb-8412-0aae81b6e6dd.png)

Original, db.py line 2225:

```python
        return json.dumps(value, default=repr)
```

Fix, db.py line 2225:

```python
        return json.dumps(value, default=repr, ensure_ascii=False)
```

PR #258, opened 2021-04-26
Fixing insert from JSON containing strings with non-ascii characters …

dylan-wu:
…are escaped aps unicode for lists, tuples, dicts

Fix of #257


simonw:
I'm going to merge this and add a test - thanks!

Issue #272, opened 2021-06-15
Idea: import CSV to memory, run SQL, export in a single command


simonw:
I quite often load a CSV file into a SQLite DB, then do stuff with it (like export results back out again as a new CSV) without any intention of keeping the CSV file around afterwards.

What if `sqlite-utils` could do this for me? Something like this:

    sqlite-utils --csv blah.csv --csv baz.csv "select * from blah join baz ..."

simonw:
Maybe also support `--csvt` as an alternative option which takes two arguments: the CSV path and the name of the table that should be created from it (rather than auto-detecting from the filename).

simonw:
How about `--json` and `--nl` and `--tsv` too? Imitating the format options for `sqlite-utils insert`.

And what happens if you provide a filename too? I'm tempted to say that the `--csv` stuff still gets loaded into an in-memory database but it's given a name and can then be joined against using SQLite `memory.blah` syntax.

simonw:
`--csvt` seems unnecessary to me: if people want to load different CSV files with the same filename (but in different directories) they will get an error unless they rename the files first.

simonw:
`--csv -` should work though, for reading from stdin. The table can be called `stdin`.

simonw:
Problem: `--csv` and `--json` and `--nl` are already options for `sqlite-utils query` - need new non-conflicting names.

simonw:
`--load-csv` and `--load-json` and `--load-nl` and `--load-tsv` are unambiguous.

eyeseast:
So, I do things like this a lot, too. I like the idea of piping in from stdin. Something like this would be nice to do in a makefile:

```sh
cat file.csv | sqlite-utils --csv --table data - 'SELECT * FROM data WHERE col="whatever"' > filtered.csv
```

If you assumed that you're always piping out the same format you're piping in, the option names don't have to change. Depends how much you want to change formats.

simonw:
This is going to need to be a separate command, for relatively non-obvious reasons.

    sqlite-utils blah.db "select * from x"

Is equivalent to this, because `query` is the default sub-command:

    sqlite-utils query blah.db "select * from x"

But... this means that making the filename optional doesn't actually work - because then this is ambiguous:

    sqlite-utils --load-csv blah.csv "select * from blah"

So instead, I'm going to add a new sub-command. I'm currently thinking `memory` to reflect that this command operates on an in-memory database:

[...]

PR #273, opened 2021-06-16
sqlite-utils memory command for directly querying CSV/JSON data

simonw:
Refs #272. Initial implementation only does CSV data, still needs:

- [ ] Use sniff to detect CSV or TSV (if `:tsv` or `:csv` was not specified) and delimiters
- [x] Implement `--save`
- [x] Add `--dump` to the documentation
- [x] Add `--attach` example to the documentation
- [ ] Add `--encoding` option
- [x] Replace `:memory:` in documentation
- [ ] Check for `[` or `{` to see if it's JSON
- [ ] Implement `filename.json:json` and `-:nl` and suchlike options for specifying the format rather than guessing it - see https://github.com/simonw/sqlite-utils/issues/272#issuecomment-861985944

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1117,6 +1117,11 @@ def query(
     help="Named :parameters for SQL query",
 )
 @click.option("--dump", is_flag=True, help="Dump SQL for in-memory database")
+@click.option(
+    "--save",
+    type=click.Path(file_okay=True, dir_okay=False, allow_dash=False),
+    help="Save in-memory database to this file",
+)
 @load_extension_option
 def memory(
     paths,
@@ -1133,6 +1138,7 @@ def memory(
     raw,
     param,
     dump,
+    save,
     load_extension,
 ):
     "Execute SQL query against an in-memory database, optionally populated by imported data"
@@ -1157,6 +1163,12 @@ def memory(
     if dump:
         for line in db.conn.iterdump():
             click.echo(line)
+        return
+
+    if save:
+        db2 = sqlite_utils.Database(save)
+        for line in db.conn.iterdump():
+            db2.execute(line)
         return
 
     for alias, attach_path in attach:
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -694,11 +694,11 @@ def insert_upsert_implementation(
         raise click.ClickException("Use just one of --nl, --csv or --tsv")
     if encoding and not (csv or tsv):
         raise click.ClickException("--encoding must be used with --csv or --tsv")
+    if pk and len(pk) == 1:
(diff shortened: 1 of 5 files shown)


simonw:
Initial documentation is here: https://github.com/simonw/sqlite-utils/blob/sqlite-utils-memory/docs/cli.rst#running-queries-directly-against-csv-data

It only talks about CSV at the moment - needs to be updated to mention JSON too once that is implemented.

sqlite_utils/cli.py
```diff
@@ -1087,8 +1087,96 @@ def query(
...
+        if i == 0:
+            view_names.append("t")
+        for view_name in view_names:
+            if not db[view_name].exists():
+                db.create_view(view_name, "select * from [{}]".format(csv_table))
+
+    if dump:
+        for line in db.conn.iterdump():
+            click.echo(line)
+        return
```
simonw, line 1160: This needs mentioning in the documentation too.

sqlite_utils/cli.py
```diff
@@ -1087,8 +1087,96 @@ def query(
...
+
+    for alias, attach_path in attach:
+        db.attach(alias, attach_path)
```
simonw, line 1163: Documentation should also mention `--attach` and show an example - it's neat to be able to e.g. pipe in a bunch of CSV data and then use that to query against an existing table.

simonw:
I can stop promoting `:memory:` here and promote `memory` instead:

https://github.com/simonw/sqlite-utils/blob/c7234cae8336b8525034e8f917d82dd0699abd42/docs/cli.rst#L83-L86

simonw:
I should probably refactor the CSV/JSON/loading stuff into a function in `utils.py` in order to share some of the implementation with the existing `sqlite-utils insert` code: https://github.com/simonw/sqlite-utils/blob/287cdcae8908916687f2ecccc87c38549d004ac6/sqlite_utils/cli.py#L691-L734

simonw:
Can't share much code though since a bunch of that `insert` stuff is specific to that command - showing progress bars, returning errors on illegal option combinations etc.

simonw:
I'm going to add a `--encoding` option - it will affect ALL CSV input files, so if you have CSV files with different encodings you'll need to sort that mess out yourself (likely by importing each CSV file separately into a database using `sqlite-utils insert` with different `--encoding` values).

simonw:
`--attach` documentation: https://github.com/simonw/sqlite-utils/blob/192dc2c5b73bd836ab8e2e5fed4b36c6ea02f250/docs/cli.rst#joining-in-memory-data-against-existing-databases-using-attach

simonw:
I'm going to merge this as-is and work on the JSON/TSV support in a separate issue.

Issue #319, opened 2021-08-22
[Enhancement] Please allow 'insert-files' to insert content as text.


pjamargh:
'insert-files' creates BLOB columns for file contents. Transforming the column to TEXT still keep the content as binary. Even though I'm sure there is a transform that can be applied decoding the text it would be great to have a argument to make 'insert-files' to do it as text (with optional text encoding).

The use case is a bunch of htmls (single file) on a directory structure that inserted with this command could be served in Datasette allowing full text search.

simonw:
This is a good idea.

simonw:
Will need to re-title this section of the documentation: https://sqlite-utils.datasette.io/en/3.16/cli.html#inserting-binary-data-from-files - "Inserting binary data from files" will become "Inserting data from files"

I'm OK with keeping the default as `BLOB` but I could add a `--text` option which stores the content as text instead.

If the text can't be stored as `utf-8` I'll probably raise an error.

pjamargh:
I was thinking that an approach could be making FILE_COLUMNS a generator (_get_file_columns(mode)) or you can just have a different set of columns (is there something else that makes sense to be changed on the text scenario?).

About UTF-8 I was referring to the encoding to use when reading files. This can be difficult to auto-detect but I believe that UTF-8 is pretty much the standard for text files.

simonw:
I'm going to assume utf-8 but allow `--encoding` to be used to specify something different, since that option is already supported by other commands.

simonw:
Here's the error message I have working for invalid unicode:
```
sqlite-utils insert-files /tmp/text.db files *.txt --text
  [------------------------------------]    0%
Error: Could not read file '/Users/simon/Dropbox/Development/sqlite-utils/data.txt' as text

'utf-8' codec can't decode byte 0xe3 in position 83: invalid continuation byte
```

pjamargh:
Oh, I misread. Yes some files will not be valid UTF-8, I'd throw a warning and continue (not adding that file) but if you want to get more elaborate you could allow to define a policy on what to do. Not adding the file, index binary content or use a conversion policy like the ones available on Python's decode.

From https://stackoverflow.com/questions/24616678/unicodedecodeerror-in-python-when-reading-a-file-how-to-ignore-the-error-and-ju :
 - 'ignore' ignores errors. Note that ignoring encoding errors can lead to data loss.
 - 'replace' causes a replacement marker (such as '?') to be inserted where there is malformed data.
[...]

PR #321, opened 2021-08-24
Ability to insert file contents as text, in addition to blob

simonw:
Refs #319.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -909,12 +909,10 @@ The command will fail if you reference columns that do not exist on the table. T
 
 .. _cli_insert_files:
 
-Inserting binary data from files
-================================
+Inserting data from files
+=========================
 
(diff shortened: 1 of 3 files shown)


simonw:
Huh, tests are failing but only on Windows!

simonw:
Hah, the error here is actually:
```
>           assert result.exit_code == 1, result.output
E           AssertionError: 
E             
E           assert 0 == 1
E            +  where 0 = <Result okay>.exit_code
```
So I was expecting an error, but instead the command worked.

I suspect this is because on Windows the default character set may not be UTF-8?

simonw:
https://discuss.python.org/t/pep-597-use-utf-8-for-default-text-file-encoding/1819 says:

> Currently, `TextIOWrapper` uses `locale.getpreferredencoding(False)` (hereinafter called “locale encoding”) when encoding is not specified.
> ...
> Package authors using macOS or Linux may forget that the default encoding
is not always UTF-8.

simonw:
I'm going to skip this test on windows.

PR #322, opened 2021-08-25
Add dict type to be mapped as TEXT in sqllite

minaeid90:
the library deal with Postgres type jsonb as dictionary, add dict type as a TEXT for mapping to sqlite

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -167,6 +167,7 @@ COLUMN_TYPE_MAPPING = {
     int: "INTEGER",
     bool: "INTEGER",
     str: "TEXT",
+    dict: "TEXT",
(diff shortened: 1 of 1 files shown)


simonw:
This relates to the fact that dictionaries, lists and tuples get special treatment and are converted to JSON strings, using this code: https://github.com/simonw/sqlite-utils/blob/e8d958109ee290cfa1b44ef7a39629bb50ab673e/sqlite_utils/db.py#L2937-L2947

So the `COLUMN_TYPE_MAPPING` should include those  too - right now it looks like this: https://github.com/simonw/sqlite-utils/blob/e8d958109ee290cfa1b44ef7a39629bb50ab673e/sqlite_utils/db.py#L165-L188

Issue #248, opened 2021-03-20
support for Apache Arrow / parquet files I/O


mhalle:
I just started looking at Apache Arrow using pyarrow for import and export of tabular datasets, and it looks quite compelling. It might be worth looking at for sqlite-utils and/or datasette.

As a test, I took a random jsonl data dump of a dataset I have with floats, strings, and ints and converted it to arrow's parquet format using the naive `pyarrow.parquet.write_file()` command, which has automatic type inferrence. It compressed down to 7% of the original size. Conversion of a 26MB JSON file and serializing it to parquet was eyeblink instantaneous. Parquet files are portable and can be directly imported into pandas and other analytics software. 

The only hangup is the automatic type inference of the naive reader. It's great for general laziness and for parsing JSON columns (it correctly interpreted a table of mine with a JSON array). However, I did get an exception for a string column where most entries looked integer-like but had a couple values that weren't -- the reader tried to coerce all of them for some reason, even though the JSON type is string. Since the writer optionally takes a schema, it shouldn't be too hard to grab the sqlite header types. With some additional hinting, you might get datetime columns and JSON, which are native Arrow types. 

Somewhat tangentially, someone even wrote an sqlite vfs extension for Parquet: https://cldellow.com/2018/06/22/sqlite-parquet-vtable.html

simonw:
Thtis could be done as a separate tool or `sqlite-utils` plugin but I don't want to add the necessary dependencies to `sqlite-utils` itself.

PR #333, opened 2021-10-28
Add functionality to read Parquet files.

Florents-Tselai:
I needed this for a project of mine, and I thought it'd be useful to have it in sqlite-utils (It's also mentioned in #248 ).
The current implementation works (data is read & data types are inferred correctly.
I've added a single straightforward test case, but @simonw please let me know if there are any non-obvious flags/combinations I should test too.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -665,6 +665,17 @@ The ``most_common`` and ``least_common`` columns will contain nested JSON arrays
 
 .. _cli_inserting_data:
 
+Inserting Parquet data
+======================
+
+Parquet is a columnar storage format, frequently used in the Hadoop/Spark ecosystem as well as cloud providers.
+Parquet files can be inserted via the ``--parquet`` flag.
+Here's an example::
+
+    $ sqlite-utils insert --parquet data.db data ./data.parquet
+
+Parquet files, along with the data store data types too, thus making the ``--detect-types`` flag redundant, as data types are inferred automatically (using PyArrow)
+
 Inserting JSON data
 ===================
 
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ setup(
         "click-default-group",
         "tabulate",
         "dateutils",
+        "pyarrow"
     ],
     setup_requires=["pytest-runner"],
     extras_require={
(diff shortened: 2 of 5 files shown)


simonw:
Some other recent projects (like trying to get this library to work in JupyterLite) have made me much more cautious about adding new dependencies, especially dependencies like `pyarrow` which require custom C/Rust extensions.

There are a few ways this could work though:

- Have this as an optional dependency feature - so it only works if the user installs `pyarrow` as well
- Implement this as a separate tool, `parquet-to-sqlite` - which could itself depend on `sqlite-utils`
- Add a concept of "plugins" to `sqlite-utils`, similar to how those work in Datasette: https://docs.datasette.io/en/stable/plugins.html

My favourite option is `parquet-to-sqlite` because that can be built without any additional changes to `sqlite-utils` at all!

I find the concept of plugins for `sqlite-utils` interesting. I've so far not had quite enough potential use-cases to convince me this is worthwhile (especially since it should be very easy to build out separate tools entirely), but I'm ready to be convinced that a plugin mechanism would be worthwhile.

Florents-Tselai:
Thanks for your reply @simonw . Tbh, my first attempt was actually the `parquet-to-sqlite` package but I already had Makefiles that relied on `SQLite-utils` and it was less intrusive to my workflow. Maybe I'll revisit that decision.
FYI: there's a `[sqlite-parquet-vtable](https://github.com/cldellow/sqlite-parquet-vtable)`

I don't think plugins make much sense either. Probably defeats the purpose of simplicity: simple database along with a pip-able package.

simonw:
I just remembered that there's one other place that this could fit: as a Datasette "insert" plugin.

This is vaporware at the moment, but the idea is that Datasette itself could grow a mechanism for importing data, that's driven by plugins.

Out of the box Datasette would be able to import CSV and CSV files, similar to `sqlite-utils insert ... --csv` - but plugins would then be able to add support for additional format such as GeoJSON or - in this case - Parquet.

The neat thing about having it as a Datasette plugin is that one plugin would enable three different ways of importing data:

1. Via a new `datasette insert ...` CLI option (similar to `sqlite-utils`)
2. Via a web form upload interface, where authenticated Datasette users would be able to upload files
3. Via an API interface, where files could be programatically submitted to a running Datasette server

I started fleshing out this idea quite a while ago but didn't make much concrete progress, maybe I should revisit it:

- https://github.com/simonw/datasette/issues/1160

PR #337, opened 2021-11-05
Default values for `--attach` and `--param` options

urbas:
It seems that `click` 8.x uses `None` as the default value for `multiple=True` options.

This change makes the code forward-compatible with `click` 8.x.

See this build failure for more info: https://hydra.nixos.org/build/156926608


urbas:
Smokes, it looks like there was a bug in click 8.0.2 (fixed in 8.0.3: https://github.com/pallets/click/issues/2089). Meaning this PR is not needed. Closing.

Issue #79, opened 2020-01-31
Helper methods for working with SpatiaLite


simonw:
As demonstrated by this piece of documentation, using SpatiaLite with sqlite-utils requires a fair bit of boilerplate:
https://github.com/simonw/sqlite-utils/blob/f7289174e66ae4d91d57de94bbd9d09fabf7aff4/docs/python-api.rst#L880-L909

eyeseast:
Came here to add this. I might pick it up.

Would also add a utility to create (and update and delete?) a spatial index. It's not much code but I have to look it up every time.

eyeseast:
Some proposals I'd add to sqlite-utils:

Some version of this, from [geojson-to-sqlite](https://github.com/simonw/geojson-to-sqlite/blob/main/geojson_to_sqlite/utils.py#L124-L130):

```python
def init_spatialite(db, lib):
    db.conn.enable_load_extension(True)
    db.conn.load_extension(lib)
    # Initialize SpatiaLite if not yet initialized
    if "spatial_ref_sys" in db.table_names():
        return
    db.conn.execute("select InitSpatialMetadata(1)")
```

Also a function for creating a spatial index:

```python
db.conn.execute("select CreateSpatialIndex(?, ?)", [table, "geometry"])
```

I don't know the nuances of updating a spatial index, or checking if one already exists. This could be a CLI method like:

```sh
sqlite-utils spatial-index spatial.db table-name column-name
```

eyeseast:
One more thing I'm going to add: A method to add a geometry column, which I'll need to do to create a spatial index on a table.

eyeseast:
@simonw I have a PR here https://github.com/simonw/sqlite-utils/pull/385 that adds Spatialite helpers on the Python side. Please let me know how it looks.

eyeseast:
Taking part of the conversation from #385 here.

> Would sqlite-utils add-geometry-column ... be a good CLI enhancement. for example?

Yes. And also `sqlite-utils create-spatial-index` would be great to have. My plan would be to add those once the Python API is settled.

simonw:
Documentation: https://sqlite-utils.datasette.io/en/latest/python-api.html#spatialite-helpers

simonw:
Shipped in 3.23: https://sqlite-utils.datasette.io/en/stable/changelog.html#v3-23

PR #385, opened 2022-01-14
Add new spatialite helper methods

eyeseast:
Refs #79 

This PR adds three new Spatialite-related methods to Database and Table:

- `Database.init_spatialite` loads the Spatialite extension and initializes it
- `Table.add_geometry_column` adds a geometry column
- `Table.create_spatial_index` creates a spatial index

Has tests and documentation. Feedback very welcome.


eyeseast:
@simonw Not sure if you've seen this, but any chance you can run the tests?

simonw:
Sorry had missed this - tests should run now.

eyeseast:
Fixed my spelling. That's a useful thing.

simonw:
`from sqlite_utils.utils import find_spatialite` is part of the documented API already:

https://sqlite-utils.datasette.io/en/3.22.1/python-api.html#finding-spatialite

To avoid needing to bump the major version number to 4 to indicate a backwards incompatible change, we should keep a `from .gis import find_spatialite` line at the top of `utils.py` such that any existing code with that documented import continues to work.

simonw:
What do you think about adding these as methods on the `Database` class instead? Then you could do:

```python
# This is with an optional argument, which if omitted runs find_spatialite() for you:
db.init_spatialite()

# Instead of:
init_spatialite(db, find_spatialite())
```
Likewise, the `add_geometry_column` and `create_spatial_index` methods could live on `Table`:
```python
# Instead of this:
add_geometry_column(db["locations"], "POINT", "geometry")
create_spatial_index(db["locations"], "geometry")

# Could have this:
db["locations"].add_geometry_column("POINT")
db["locations"].create_spatial_index("geometry")
```
On the one hand, this is much more consistent with the existing `sqlite-utils` Python API.

But on the other hand... this is mixing SpatiaLite functionality directly into the core classes. Is that a good idea, seeing as SpatiaLite is both an optional extension (which can be tricky to install) AND something that has a very different release cadence and quality-of-documentation from SQLite itself?

There's a third option: the SpatiaLite could exist on subclasses of `Database` and `Table` - so the above examples would look something like this:

```python
from sqlite_utils.gis import SpatiaLiteDatabase

db = SpatiaLiteDatabase("geo.db")
db.init_spatialite()
db["locations"].add_geometry_column("POINT")
db["locations"].create_spatial_index("geometry")
```

On the one hand, this would keep the SpatiaLite-specific stuff out of the core Database/Table classes. But it feels a bit untidy to me, especially since it raises the spectre of someone who was already subclassing Database for some reason now needing to instead subclass `SpatiaLiteDatabase` (not too keen on that capitalization) - or even (horror) trying to dabble with multiple inheritance, which can only lead to pain.

So I don't have a strong opinion formed on this question yet!

simonw:
I'm not sure I like `name="geometry"` as the default argument to `add_geometry_column` - mainly because of this example here:
```python
add_geometry_column(db["locations"], "POINT")
create_spatial_index(db["locations"], "geometry")
```
I had to go and look at the code to figure out if `"POINT"` was the name of the column - and I don't like how inconsistent it looks next to the following `create_spatial_index()` call where you DO need to pass the column name.

sqlite_utils/gis.py
```diff
@@ -0,0 +1,144 @@
...
+    if "spatial_ref_sys" in db.table_names():
+        return False
+    cursor = db.execute("select InitSpatialMetadata(1)")
+    result = cursor.fetchone()
+    return result and bool(result[0])
+
+
+def add_geometry_column(
+    table: Table,
+    geometry_type: str,
```
simonw, line 73: I'm trying to figure out the list of valid geometry type values, which is surprisingly hard!

I eventually found a list on http://www.gaia-gis.it/gaia-sins/spatialite-sql-5.0.1.html by searching for `AddGeometryColumn` - that documentation says that valid values are:

- `POINT`, `POINTZ`, `POINTM`, `POINTZM`
- `LINESTRING`, `LINESTRINGZ`, `LINESTRINGM`, `LINESTRINGZM`
- `POLYGON`, `POLYGONZ`, `POLYGONM`, `POLYGONZM`
- `MULTIPOINT`, `MULTIPOINTZ`, `MULTIPOINTM`, `MULTIPOINTZM`
- `MULTILINESTRING`, `MULTILINESTRINGZ`, `MULTILINESTRINGM`, `MULTILINESTRINGZM`
- `MULTIPOLYGON`, `MULTIPOLYGONZ`, `MULTIPOLYGONM`, `MULTIPOLYGONZM`
- `GEOMETRYCOLLECTION`, `GEOMETRYCOLLECTIONZ`, `GEOMETRYCOLLECTIONZM`, `GEOMETRYCOLLECTIONZM`
- `GEOMETRY`, `GEOMETRYZ`, `GEOMETRYM`, `GEOMETRYZM`

I was going to suggest maybe providing constants for these and requiring that the argument was one of those constants... but with that many options I'm not sure that would be better than sticking with strings, and letting SpatiaLite itself raise an exception on an invalid value!

simonw: I wonder if `GEOMETRY` would be a reasonable default value for `add_geometry_column(geometry_type=)`? I believe that's a column which can hold any type of geometry value - then we could allow users who want to be more restrictive to use `geometry_type="POINT"` or similar.

eyeseast:
I thought about adding these as methods on `Database` and `Table`, and I'm back and forth on it for the same reasons you are. It's certainly cleaner, and it's clearer what you're operating on. I could go either way. 

I do sort of like having all the Spatialite stuff in its own module, just because it's built around an extension you might not have or want, but I don't know if that's a good reason to have a different API.

You could have `init_spatialite` add methods to `Database` and `Table`, so they're only there if you have Spatialite set up. Is that too clever? It feels too clever.

sqlite_utils/gis.py
eyeseast: Yeah, I had to dig for that, too. And those types are case-insensitive, as far as I can tell. I'm happy to provide constants, but if you pass in `"point"` as your geometry type, Spatialite won't complain.

eyeseast: I'm fine to take out the default column name. Using `geometry` was my effort to standardize on GeoJSON, but I've seen plenty of examples that don't. Then you could have a similar signature, something like this:

```python
add_geometry_column(db["locations"], "geometry", "POINT")
create_spatial_index(db["locations"], "geometry")
```

eyeseast:
> `from sqlite_utils.utils import find_spatialite` is part of the documented API already:
> 
> https://sqlite-utils.datasette.io/en/3.22.1/python-api.html#finding-spatialite
> 
> To avoid needing to bump the major version number to 4 to indicate a backwards incompatible change, we should keep a `from .gis import find_spatialite` line at the top of `utils.py` such that any existing code with that documented import continues to work.

This is fixed now. I had to take out the type annotations for `Database` and `Table` to avoid a circular import, but that's fine and may be moot if these become class methods.

simonw:
> I thought about adding these as methods on `Database` and `Table`, and I'm back and forth on it for the same reasons you are. It's certainly cleaner, and it's clearer what you're operating on. I could go either way.
> 
> I do sort of like having all the Spatialite stuff in its own module, just because it's built around an extension you might not have or want, but I don't know if that's a good reason to have a different API.
> 
> You could have `init_spatialite` add methods to `Database` and `Table`, so they're only there if you have Spatialite set up. Is that too clever? It feels too clever.

Yeah that's too clever. You know what? I'm pretty confident we are both massively over-thinking this. We should put the methods on `Database` and `Table`! API simplicity and consistency matters more than vague concerns about purity.

eyeseast:
Works for me. I was just looking at how the FTS extensions work and they're just methods, too. So this can be consistent with that.

eyeseast:
OK, I moved all the GIS helpers into `db.py` as methods on `Database` and `Table`, and I put `find_spatialite` back in `utils.py`. I deleted `gis.py`, since there's nothing left it. Docs and tests are updated and passing.

I think this is better.

sqlite_utils/db.py
```diff
@@ -3011,6 +3048,85 @@ def truncate(value):
...
+            ],
+        )
+
+        result = cursor.fetchone()
+        return result and bool(result[0])
+
+    def create_spatial_index(self, column_name) -> bool:
+        """
+        A spatial index allows for significantly faster bounding box queries.
+        To create one, use ``create_spatial_index`` with a :ref:`table <reference_db_table>`
```
simonw, line 3100: I think this documentation is out of date, since you don't need to pass a table any more.

sqlite_utils/db.py
```diff
@@ -930,6 +930,43 @@ def analyze(self, name=None):
             sql += " [{}]".format(name)
         self.execute(sql)
 
+    def init_spatialite(self, path: str) -> bool:
```
simonw, line 933: I still like the idea of making `path` here optional - if you don't provide it (or set `path=None`) then the function can call `find_spatialite()` for you.

sqlite_utils/db.py
```diff
@@ -3011,6 +3048,85 @@ def truncate(value):
...
+    def add_geometry_column(
+        self,
+        column_name: str,
+        geometry_type: str,
+        srid: int = 4326,
+        coord_dimension: str = "XY",
+        not_null: bool = False,
+    ) -> bool:
+        """
+        In Spatialite, a geometry column can only be added to an existing table.
```
simonw, line 3060: Tiny style thing: `Spatialite` here should be `SpatiaLite` for consistency with other documentation.

sqlite_utils/db.py
```diff
@@ -3011,6 +3048,85 @@ def truncate(value):
...
+        To do so, use ``table.add_geometry_column``, passing in a geometry type.
+
+        By default, this will add a nullable column called ``geometry`` using
```
simonw, line 3063: The docs here say that the default will be a "geometry" but line 3054 doesn't yet set that default.

tests/test_gis.py
```diff
@@ -0,0 +1,98 @@
...
+    table = db.create_table("locations", {"id": str, "properties": str})
+    assert table.add_geometry_column("geometry", "Point")
+
+    # index it
+    assert table.create_spatial_index("geometry")
+
+    assert "idx_locations_geometry" in db.table_names()
+
+
+@pytest.mark.skipif(not find_spatialite(), reason="Could not find SpatiaLite extension")
```
simonw, line 78: Might be neat to refactor these checks out into a custom decorator, since we have to duplicate them for each of these tests.

Could even apply it to the entire module, as seen in this example:  https://github.com/simonw/s3-credentials/blob/6c0a34763de90fdd9448157ddecf8d26275fe791/tests/test_integration.py#L15-L16

eyeseast: That's a neat trick.

sqlite_utils/db.py
eyeseast: Fixed.

sqlite_utils/db.py
eyeseast: Got it.

sqlite_utils/db.py
eyeseast: Done.

sqlite_utils/db.py
eyeseast: Updated.

simonw:
This looks fantastic, thanks for all of the work you put into this!

simonw:
Shipped this as `sqlite-utils` 3.23: https://sqlite-utils.datasette.io/en/stable/changelog.html#v3-23

eyeseast:
Awesome. Thanks for your help getting it in. Will now look at adding CLI versions of this. It's going to be super helpful on a bunch of my projects.

PR #407, opened 2022-02-15
Add SpatiaLite helpers to CLI

### LINKED ISSUE #398: Add SpatiaLite helpers to CLI
Now that #385 is merged, add CLI versions of those methods.

```sh
# init spatialite
sqlite-utils init-spatialite database.db

# or maybe/also
sqlite-utils create database.db --enable-wal --spatialite

# add geometry columns
# needs a database, table, geometry column name, type, with optional SRID and not-null
# this needs to create a table if it doesn't already exist
sqlite-utils add-geometry-column database.db table-name geometry --srid 4326 --not-null

# spatial index an existing table/column
sqlite-utils create-spatial-index database.db table-name geometry
```

Should be mostly straightforward. The one thing worth highlighting in docs is that geometry columns can only be added to existing tables. Trying to add a geometry column to a table that doesn't exist yet might mean you have a schema like `{"rowid": int, "geometry": bytes}`. Might be worth nudging people to explicitly create a table first, then add geometry columns.

eyeseast:
Closes #398 

This adds SpatiaLite helpers to the CLI. 

```sh
# init spatialite when creating a database
sqlite-utils create database.db --enable-wal --init-spatialite

# add geometry columns
# needs a database, table, geometry column name, type, with optional SRID and not-null
# this will throw an error if the table doesn't already exist
sqlite-utils add-geometry-column database.db table-name geometry --srid 4326 --not-null

# spatial index an existing table/column
# this will throw an error it the table and column don't exist
sqlite-utils create-spatial-index database.db table-name geometry
```

Docs and tests are included.

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -2746,10 +2746,24 @@ To load it from a specific path, use --load-extension.""",
         case_sensitive=False,
     ),
     default="GEOMETRY",
+    help="Specify a geometry type for this column.",
+    show_default=True,
 )
-@click.option("--srid", type=int, default=4326)
-@click.option("--dimensions", "coord_dimension", type=str, default="XY")
-@click.option("--not-null", "not_null", is_flag=True)
+@click.option(
+    "--srid",
+    type=int,
+    default=4326,
+    show_default=True,
+    help="Spatial Reference ID. See https://spatialreference.org for details on specific projections.",
+)
+@click.option(
+    "--dimensions",
+    "coord_dimension",
+    type=str,
+    default="XY",
+    help="Coordinate dimensions. Use XYZ for three-dimensional geometries.",
+)
+@click.option("--not-null", "not_null", is_flag=True, help="Add a NOT NULL constraint.")
 @load_extension_option
 def add_geometry_column(
     db_path,
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -2810,7 +2810,7 @@ def create_spatial_index(db_path, table, column_name, load_extension):
         raise click.ClickException(
             "You must add a geometry column before creating a spatial index"
         )
-    
+
     db[table].create_spatial_index(column_name)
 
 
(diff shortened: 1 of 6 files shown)


eyeseast:
@simonw I think this is ready for a look.

sqlite_utils/cli.py
```diff
@@ -1374,6 +1378,15 @@ def create_database(path, enable_wal):
...
     if enable_wal:
         db.enable_wal()
+
+    # load spatialite from expected locations
+    if init_spatialite:
+        db.init_spatialite()
+
+    # load spatialite or another extension from a custom location
+    if load_extension:
+        _load_extensions(db, load_extension)
```
simonw, line 1388: I think this needs to happen before the call to `db.init_spatialite()` to ensure SpatiaLite is correctly loaded in the case where it wasn't in an expected location so the user had to use `--load-extension` with it.

sqlite_utils/cli.py
```diff
@@ -2701,6 +2714,120 @@ def convert(
             )
 
 
+@cli.command(
+    "add-geometry-column",
+    help="""Add a SpatiaLite geometry column to an existing table. Requires SpatiaLite extension.
```
simonw, line 2719: Any reason you used `help=` here rather than having this as a docstring on the comment? Asking because it's inconsistent with the other CLI functions in this module.

simonw:
We should add SpatiaLite to the action that calculates code coverage - that way we can calculate coverage across the new GIS tests as well: https://github.com/simonw/sqlite-utils/blob/main/.github/workflows/test-coverage.yml

Should just be a case of adding this to that workflow:

```
    - name: Install SpatiaLite
      run: sudo apt-get install libsqlite3-mod-spatialite
```

simonw:
Wow, just found out I can edit files in this PR branch by hitting `.` on my keyboard while looking at the PR, then making changes in the VS Code for web on `github.dev`!

sqlite_utils/cli.py
eyeseast: Switched. Good catch.

eyeseast:
Happy to do it and have it in the library. Going to use it a bunch. This whole SpatiaLite toolchain become a huge part of my work in the past year.

PR #429, opened 2022-05-02
Depend on click-default-group-wheel

simonw:
Trying to get this to work with Pyodide.

Refs: https://github.com/simonw/click-default-group-wheel/issues/3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     install_requires=[
         "sqlite-fts4",
         "click",
-        "click-default-group",
+        "click-default-group-wheel",
         "tabulate",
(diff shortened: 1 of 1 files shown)


simonw:
I made a build of this branch and tested it like this: https://pyodide.org/en/stable/console.html

```
>>> import micropip
>>> await micropip.install("https://s3.amazonaws.com/simonwillison-cors-allowed-public/sqlite_utils-3.26-py3-none-any.whl")
>>> import sqlite_utils
>>> list(db.query("select 32443 + 55"))
[{'32443 + 55': 32498}]
```

Issue #449, opened 2022-06-22
Utilities for duplicating tables and creating a table with the results of a query


davidleejy:
is there a duplicate table functionality? Otherwise, I'd be happy to submit a PR.

In sqlite3 it would look like:

```python
import sqlite3 as sl

con = sl.connect('prompt-tune.db')

def db_duplicate_table(table_name, table_name_new, con=con):
    # Duplicates table `table_name` to a new table `table_name_new`.
    try:
        cur = con.cursor()
        cur.execute(f"""CREATE TABLE {table_name_new} AS SELECT * FROM {table_name}""")
    except Exception as e:
        print(e)
    finally:
        cur.close()

db_duplicate_table('orig_table', 'new_table')
```

simonw:
I like the idea of this as a feature - design could look like this:

    db["my_table"].duplicate("new_table")

Then for the CLI tool:

    sqlite-utils duplicate data.db my_table new_table

Additional features that might be useful:

- Duplicate across connection aliases - so you can copy a table to another database using the `duplicate` method or command
- Ability to execute an arbitrary SQL query and save the results to a new table - basically a wrapper around `CREATE TABLE t AS SELECT ...`, again with alias support for copying to another database

davidleejy:
I've written the code and test. Would you be able to advise how to compare table columns in a pytest function?

Test:
```python
def test_duplicate(fresh_db):
    table = fresh_db.create_table(
        "table1",
        {
            "text_col": str,
            "float_col": float,
            "int_col": int,
            "bool_col": bool,
            "bytes_col": bytes,
            "datetime_col": datetime.datetime,
        },
    )
    dt = datetime.datetime.now()
    b = bytes('hello world', 'utf-8')
    data = {"text_col": "Cleo", 
            "float_col": 3.14,
            "int_col": -2,
            "bool_col": True,
            "bytes_col": b,
            "datetime_col": str(dt)}
    table1 = fresh_db["table1"]
    row_id = table1.insert(data).last_rowid
    table1.duplicate('table2')
    table2 = fresh_db["table2"]
    assert data == table2.get(row_id)
    assert table1.columns == table2.columns    # Fails here
```

Result:
![Screenshot 2022-07-05 at 1 31 55 AM](https://user-images.githubusercontent.com/1690072/177198814-daac48c9-5746-49d0-a14a-14fe181c5a2f.png)

[...]

PR #452, opened 2022-07-09
Add duplicate table feature

davidleejy:
This PR addresses a feature request raised in issue #449. Specifically this PR adds a functionality that lets users duplicate a table via:

```python
table_new = db["my_table"].duplicate("new_table")
```

Test added in file `tests/test_duplicate.py`.

Happy to make changes to meet maintainers' feedback, if any.

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -305,11 +305,7 @@ class Database:
         ), "Either specify a filename_or_conn or pass memory=True"
         if memory_name:
             uri = "file:{}?mode=memory&cache=shared".format(memory_name)
-            self.conn = sqlite3.connect(
-                uri,
-                uri=True,
-                check_same_thread=False,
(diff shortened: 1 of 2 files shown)


sqlite_utils/db.py
```diff
@@ -1476,6 +1476,21 @@ def create(
...
         return self
 
+    def duplicate(self, name_new: str) -> "Table":
+        """
+        Duplicate this table in this database.
+
+        :param name_new: Name of new table.
+        """
+        assert self.exists()
+        with self.db.conn:
```
simonw, line 1486: I questioned if running a transaction here was the right thing to do - what if the user wants to manage their own transactions? But then I saw that the `.create()` method does the same thing, and since `.duplicate()` is a similar operation to `.create()` I think this is the right approach (consistency is really important).

PR #455, opened 2022-07-19
in extract code, check equality with IS instead of = for nulls

### LINKED ISSUE #423: .extract() doesn't set foreign key when extracted columns contain NULL value
I've run into an issue with `extract` and I don't believe this is the intended behaviour.

I'm working with a database with music listening information. Currently it has one large table `listens` that contains all information. I'm trying to normalize the database by extracting relevant columns to separate tables (`artists`, `tracks`, `albums`). Not every track has an album.

A simplified demonstration with just `track_title` and `album_title` columns:
```ipython
In [1]: import sqlite_utils

In [2]: db = sqlite_utils.Database(memory=True)

In [3]: db["listens"].insert_all([
   ...:     {"id": 1, "track_title": "foo", "album_title": "bar"},
   ...:     {"id": 2, "track_title": "baz", "album_title": None}
   ...: ], pk="id")
Out[3]: <Table listens (id, track_title, album_title)>
```

The track in the first row has an album, the second track doesn't. Now I extract album information into a separate column:
```ipython
In [4]: db["listens"].extract(columns=["album_title"], table="albums", fk_column="album_id")
Out[4]: <Table listens (id, track_title, album_id)>

In [5]: list(db["albums"].rows)
Out[5]: [{'id': 1, 'album_title': 'bar'}, {'id': 2, 'album_title': None}]

In [6]: list(db["listens"].rows)
Out[6]: 
[{'id': 1, 'track_title': 'foo', 'album_id': 1},
 {'id': 2, 'track_title': 'baz', 'album_id': None}]
```

This behaves as expected -- the `album` table contains entries for both the existing album and the NULL album. The `listens` table has a foreign key only for the first row (since the album in the second row was empty).

Now I want to extract the track information as well. Album information belongs to the track so I want to extract both columns to a new table.
```ipython
In [7]: db["listens"].extract(columns=["track_title", "album_id"], table="tracks", fk_column="track_id")
Out[7]: <Table listens (id, track_id)>

In [8]: list(db["tracks"].rows)
Out[8]: 
[{'id': 1, 'track_title': 'foo', 'album_id': 1},
 {'id': 2, 'track_title': 'baz', 'album_id': None}]

In [9]: list(db["listens"].rows)
Out[9]: [{'id': 1, 'track_id': 1}, {'id': 2, 'track_id': None}]
```

Extracting to the `tracks` table worked fine (both tracks are present with correct columns). However, the `listens` table only has a foreign key to the newly created tracks for the first row, the foreign key in the second row is NULL.

Changing the order of extracts doesn't help.

I poked around in the source a bit and I believe [this line](https://github.com/simonw/sqlite-utils/blob/433813612ff9b4b501739fd7543bef0040dd51fe/sqlite_utils/db.py#L1737) (essentially comparing `NULL = NULL`) is the problem, but I don't know enough about SQL to create a reliable fix myself.

fgregg:
sqlite "IS" is equivalent to SQL "IS NOT DISTINCT FROM"

closes #423

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -1761,7 +1761,7 @@ class Table(Queryable):
                 magic_lookup_column=magic_lookup_column,
                 lookup_table=table,
                 where=" AND ".join(
-                    "[{table}].[{column}] = [{lookup_table}].[{lookup_column}]".format(
+                    "[{table}].[{column}] IS [{lookup_table}].[{lookup_column}]".format(
                         table=self.name,
                         lookup_table=table,
                         column=column,


simonw:
Can you add a new test for this? Something derived from the example in #423 would work great.

simonw:
I did not know `is` worked against strings and integers and booleans!

https://latest.datasette.io/_memory?sql=select+1+is+1%2C+2.3+is+2.3%2C+true+is+true%2C+true+is+false%2C+%27dog%27+is+%27dog%27%2C+%27dog%27+is+%27cat%27%2C+null+is+null%2C+null+is+not+null

```sql
select 1 is 1, 2.3 is 2.3, true is true, true is false, 'dog' is 'dog', 'dog' is 'cat', null is null, null is not null
```

|   1 is 1 |   2.3 is 2.3 |   true is true |   true is false |   'dog' is 'dog' |   'dog' is 'cat' |   null is null |   null is not null |
|----------|--------------|----------------|-----------------|------------------|------------------|----------------|--------------------|
|        1 |            1 |              1 |               0 |                1 |                0 |              1 |                  0 |

Issue #467, opened 2022-08-23
Mechanism for ensuring a table has all the columns


simonw:
Suggested by @jefftriplett on Discord: https://discord.com/channels/823971286308356157/997738192360964156/1011655389063958600

simonw:
Jeff suggested `db.create(table_name, ..., alter=True)` for this.

simonw:
I'm not crazy about having to pass both `alter=True` and `if_not_exists=True` - maybe `alter` should imply `if_not_exists`.

simonw:
Should passing `alter=True` also drop any columns that aren't included in the new table structure?

It could even spot column types that aren't correct and fix those.

Is that consistent with the expectations set by how `alter=True` works elsewhere?

simonw:
Could call it `ensure=True` here if it works differently enough from `alter=True` that the behavior could be confusing.

simonw:
Thinking about this more, I think `alter=True` is a good name for this option even if it does more than the same option on `.insert()`.

simonw:
Maybe there should be a separate `table.alter(...)` method that does the actual work here, with `.create(..., alter=True)` as syntactic sugar for triggering that if the table exists already.

simonw:
... but that's what the `table.transform(...)` method does already!

So maybe this is actually a `transform=True` parameter to `create()` that triggers `table.transform(...)` if necessary.

jefftriplett:
> Should passing `alter=True` also drop any columns that aren't included in the new table structure?
> 
> It could even spot column types that aren't correct and fix those.
> 
> Is that consistent with the expectations set by how `alter=True` works elsewhere?

I would lean towards not dropping them (or making a `drop=True` or `drop_columns=True`or `drop_missing_columns=True`) to work with existing tables easier. 

I do like that sqlite-utils mostly just works with existing tables but it's also nice to add to existing fields in a few cases.

simonw:
Initial prototype:
```diff
diff --git a/sqlite_utils/db.py b/sqlite_utils/db.py
index 18a442a..03fd345 100644
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -875,6 +875,7 @@ class Database:
         hash_id_columns: Optional[Iterable[str]] = None,
         extracts: Optional[Union[Dict[str, str], List[str]]] = None,
         if_not_exists: bool = False,
+        transform: bool = False,
     ) -> "Table":
         """
[...]

PR #468, opened 2022-08-23
db[table].create(..., transform=True) and create-table --transform

simonw:
Work in progress. Still needs documentation and tests (and to cover more cases of things that might have changed).

Refs:
- #467

----
:books: Documentation preview :books:: https://sqlite-utils--468.org.readthedocs.build/en/468/

### DIFF
--- a/docs/cli-reference.rst
+++ b/docs/cli-reference.rst
@@ -856,6 +856,7 @@ See :ref:`cli_create_table`.
                                 foreign key
       --ignore                  If table already exists, do nothing
       --replace                 If table already exists, replace it
+      --transform               If table already exists, try to transform the schema
       --load-extension TEXT     SQLite extensions to load
       -h, --help                Show this message and exit.
 
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1520,6 +1520,8 @@ You can specify foreign key relationships between the tables you are creating us
 
 If a table with the same name already exists, you will get an error. You can choose to silently ignore this error with ``--ignore``, or you can replace the existing table with a new, empty table using ``--replace``.
 
+You can also pass ``--transform`` to transform the existing table to match the new schema. See :ref:`python_api_explicit_create` in the Python library documentation for details of how this option works.
+
 .. _cli_duplicate_table:
 
 Duplicating tables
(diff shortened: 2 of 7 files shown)


simonw:
The main challenge here is coming up with comprehensive tests. The cases I need to cover are from this block of code:

```python
    def create_table(
        self,
        name: str,
        columns: Dict[str, Any],
        pk: Optional[Any] = None,
        foreign_keys: Optional[ForeignKeysType] = None,
        column_order: Optional[List[str]] = None,
        not_null: Iterable[str] = None,
        defaults: Optional[Dict[str, Any]] = None,
        hash_id: Optional[str] = None,
        hash_id_columns: Optional[Iterable[str]] = None,
        extracts: Optional[Union[Dict[str, str], List[str]]] = None,
        if_not_exists: bool = False,
        transform: bool = False,
    ) -> "Table":

```

simonw:
For the moment I'm not going to pay attention to `foreign_keys=` - I will note that these are not modified in the documentation.

simonw:
Interesting challenge with `default_value`: I need to be able to tell if the default values passed to `.create()` differ from those in the database already.

Introspecting that is a bit tricky:

```pycon
>>> import sqlite_utils
>>> db["blah"].create({"id": int, "name": str}, not_null=("name",), defaults={"name": "bob"})
<Table blah (id, name)>
>>> db["blah"].columns
[Column(cid=0, name='id', type='INTEGER', notnull=0, default_value=None, is_pk=0), Column(cid=1, name='name', type='TEXT', notnull=1, default_value="'bob'", is_pk=0)]
```
Note how a default value of the Python string `bob` is represented in the results of `PRAGMA table_info()` as `default_value="'bob'"` - it's got single quotes added to it!

So comparing default values from introspecting the database needs me to first parse that syntax. This may require a new table introspection method.

simonw:
Just needs documentation now.

simonw:
Documentation:
- https://sqlite-utils--468.org.readthedocs.build/en/468/python-api.html#explicitly-creating-a-table
- https://sqlite-utils--468.org.readthedocs.build/en/468/cli.html#creating-tables

PR #480, opened 2022-08-30
search_sql add include_rank option

chapmanjacobd:
I haven't tested this yet but wanted to get a heads-up whether this kind of change would be useful or if I should just duplicate the function and tweak it within my code

----
:books: Documentation preview :books:: https://sqlite-utils--480.org.readthedocs.build/en/480/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2376,6 +2376,7 @@ class Table(Queryable):
         limit: Optional[int] = None,
         offset: Optional[int] = None,
         where: Optional[str] = None,
+        include_rank: bool = False,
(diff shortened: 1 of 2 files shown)


simonw:
Yeah this seems like a reasonable addition to me. 

Needs a test, which can go next to this one: https://github.com/simonw/sqlite-utils/blob/087753cd42c406f1e060c1822dcd9b5fda3d60f4/tests/test_fts.py#L561

chapmanjacobd:
Thanks for pointing me to the right place

simonw:
Tests look great, thank you!

PR #486, opened 2022-09-08
progressbar for inserts/upserts of all fileformats, closes #485

### LINKED ISSUE #485: Progressbar not shown when inserting/upserting jsonlines file
When inserting or upserting a jsonlines file, no progressbar is shown. Expected behavior is that, just like with .csv/.tsv files, also for a jsonlines file (--nl), unless --silent is provided, a progressbar is shown.

```bash
sql-utils upsert mydb.db posts posts.jl --nl --pk post_id
(silence)
```

Currently `file_progress` is only called within the tsv/csv logic, however I think it can be safely wrapped around all the all the input formats that use `decoded`: https://github.com/simonw/sqlite-utils/blob/main/sqlite_utils/cli.py#L963

MischaU8:
----
:books: Documentation preview :books:: https://sqlite-utils--486.org.readthedocs.build/en/486/

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,12 +9,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.6"]
         os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
     - uses: actions/cache@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,13 +10,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.6"]
         numpy: [0, 1]
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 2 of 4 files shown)


simonw:
Annoying `mypy` test failure:

```
/Users/runner/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/numpy/__init__.pyi:636:
error: Positional-only parameters are only supported in Python 3.8 and greater
```
Looks like this:
- https://github.com/python/mypy/issues/13627

simonw:
There's a fix for `mypy` that has landed but isn't out in a release yet:
- https://github.com/python/mypy/issues/13385

For the moment looks like pinning to Python 3.10.6 could help. Need to figure out how to do that in GitHub Actions though.

simonw:
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input says can set the full version:

```
- uses: actions/setup-python@v4
  with:
    python-version: "3.10.6" 
```
I'll try that.

simonw:
This looks good to me. I need to run some manual tests before merging (it's a good sign that the automated tests pass though).

simonw:
I had a big CSV file lying around, I converted it to other formats like this:

    sqlite-utils insert /tmp/t.db t /tmp/en.openfoodfacts.org.products.csv --csv
    sqlite-utils rows /tmp/t.db t --nl > /tmp/big.nl
    sqlite-utils rows /tmp/t.db t > /tmp/big.json

Then tested the progress bar like this:

    sqlite-utils insert /tmp/t2.db t /tmp/big.nl --nl

Output:

```
sqlite-utils insert /tmp/t2.db t /tmp/big.nl --nl
  [------------------------------------]    0%
  [#######-----------------------------]   20%  00:00:20
```
With `--silent` it is silent.

And for regular JSON:

```
sqlite-utils insert /tmp/t3.db t /tmp/big.json 
  [####################################]  100%
```
This is actually not doing the right thing. The problem is that `sqlite-utils` doesn't include a streaming JSON parser, so it instead reads that entire JSON file into memory first (exhausing the progress bar to 100% instantly) and then does the rest of the work in-memory while the bar sticks at 100%.

simonw:
I'm going to land this anyway. The lack of a streaming JSON parser is a separate issue, I don't think it should block landing this improvement.

PR #498, opened 2022-10-11
fix: enable-fts permanently save triggers

chapmanjacobd:
I was wondering why my all my databases were giving wild search results. Turns out create_trigger was not sticking!

Running `sqlite-utils triggers x.db` shows `[]` after running `enable-fts` using the python api. Looking at the counts trigger it seems that is the right way to save triggers. triggers show up now

----
:books: Documentation preview :books:: https://sqlite-utils--498.org.readthedocs.build/en/498/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2266,7 +2266,8 @@ class Table(Queryable):
                     new_cols=new_cols,
                 )
             )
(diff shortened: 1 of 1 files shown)


chapmanjacobd:
nevermind it was because I was running `db[table].transform`. The fts tables would still be there but the triggers would be dropped

PR #499, opened 2022-10-11
feat: recreate fts triggers after table transform

chapmanjacobd:
https://github.com/simonw/sqlite-utils/pull/498

----
:books: Documentation preview :books:: https://sqlite-utils--499.org.readthedocs.build/en/499/

alternatively, `self.disable_fts()`


chapmanjacobd:
The problem with how it is currently is that the transformed fts table will return incorrect results, even if create_triggers was enabled previously. Maybe the simplest solution is to disable fts on a transformed table rather than try to recreate it? Thoughts?

PR #508, opened 2022-10-31
Allow surrogates in parameters

### LINKED ISSUE #507: conn.execute: UnicodeEncodeError: 'utf-8' codec can't encode character
I'm not really sure what caused this and it happened in the middle of my program (after running for 35775 seconds).

```
Extracting metadata 49.9% (chunk 9893 of 19831)
...
  File "/home/xk/.local/lib/python3.10/site-packages/xklb/fs_extract.py", line 90, in extract_chunk
    args.db["media"].insert_all(utils.list_dict_filter_bool(media), pk="path", alter=True, replace=True)
  File "/home/xk/.local/lib/python3.10/site-packages/sqlite_utils/db.py", line 3107, in insert_all
    self.insert_chunk(
  File "/home/xk/.local/lib/python3.10/site-packages/sqlite_utils/db.py", line 2872, in insert_chunk
    result = self.db.execute(query, params)
  File "/home/xk/.local/lib/python3.10/site-packages/sqlite_utils/db.py", line 483, in execute
    return self.conn.execute(sql, parameters)
UnicodeEncodeError: 'utf-8' codec can't encode character '\udcc3' in position 62: surrogates not allowed
```

This might be relevant: https://stackoverflow.com/questions/31898353/python-cant-encode-with-surrogateescape

I'm going to try re-running with 

```py
    def execute(
        self, sql: str, parameters: Optional[Union[Iterable, dict]] = None
    ) -> sqlite3.Cursor:
        """
        Execute SQL query and return a ``sqlite3.Cursor``.

        :param sql: SQL query to execute
        :param parameters: Parameters to use in that query - an iterable for ``where id = ?``
          parameters, or a dictionary for ``where id = :id``
        """
        try:
            if self._tracer:
                self._tracer(sql, parameters)
            if parameters is not None:
                return self.conn.execute(sql, parameters)
            else:
                return self.conn.execute(sql)
        except UnicodeEncodeError:
            sql = sql.encode('utf-8', 'surrogatepass').decode('utf-8')
            if parameters is not None:
                parameters = parameters.encode('utf-8', 'surrogatepass').decode('utf-8')
            return self.execute(sql, parameters)
```

chapmanjacobd:
closes #507

https://dwheeler.com/essays/fixing-unix-linux-filenames.html

----
:books: Documentation preview :books:: https://sqlite-utils--508.org.readthedocs.build/en/508/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -477,12 +477,18 @@ class Database:
         :param parameters: Parameters to use in that query - an iterable for ``where id = ?``
           parameters, or a dictionary for ``where id = :id``
         """
-        if self._tracer:
-            self._tracer(sql, parameters)
-        if parameters is not None:
-            return self.conn.execute(sql, parameters)
-        else:
-            return self.conn.execute(sql)
+        try:
+            if self._tracer:
+                self._tracer(sql, parameters)
+            if parameters is not None:
+                return self.conn.execute(sql, parameters)
+            else:
+                return self.conn.execute(sql)
+        except UnicodeEncodeError:
+            sql = sql.encode('utf-8', 'surrogatepass').decode('utf-8')
+            if parameters is not None:
+                parameters = parameters.encode('utf-8', 'surrogatepass').decode('utf-8')
+            return self.execute(sql, parameters)
 
     def executescript(self, sql: str) -> sqlite3.Cursor:
         """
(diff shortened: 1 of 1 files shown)


chapmanjacobd:
Maybe this is actually a problem in the python sqlite bindings. Given [SQLITE's stance on this](https://www.sqlite.org/invalidutf.html) they should probably use `encode('utf-8', 'surrogatepass')`. As far as I understand the error here won't actually be resolved by this PR as-is. We would need to modify the data with `surrogateescape`... :/

PR #515, opened 2022-11-26
upsert new rows with constraints, fixes #514

### LINKED ISSUE #514: upsert of new row with check constraints fails
(I originally opened this in https://github.com/simonw/datasette-insert/issues/20, but I see that that library depends on sqlite-utils)

In the case of a new row, upsert first adds the row, specifying only its pkeys: https://github.com/simonw/sqlite-utils/blob/965ca0d5f5bffe06cc02cd7741344d1ddddf9d56/sqlite_utils/db.py#L2783-L2787

This means that a table with NON NULL (or other constraint) columns that aren't part of the pkey can't have new rows upserted.

cldellow:
This fixes #514 by making the initial insert for upserts include all columns, so that new rows can be added to tables with non-pkey columns that have constraints.

(aside: I'm not a python programmer. `pip`? `pipenv`? `venv`? These are mystical incantations to me. The process to set up this repo for local development and testing was _so easy_. Thank you for the excellent contributing documentation!)

----
:books: Documentation preview :books:: https://sqlite-utils--515.org.readthedocs.build/en/515/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2779,13 +2779,16 @@ class Table(Queryable):
             self.last_pk = None
             for record_values in values:
                 # TODO: make more efficient:
+
+                # The initial insert statement includes all columns, so that upserts
+                # of new rows whose non-pk columns have constraints can succeed
                 record = dict(zip(all_columns, record_values))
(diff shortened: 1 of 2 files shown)


simonw:
I should have spotted this PR before I shipped my own fix! https://github.com/simonw/sqlite-utils/commit/2376c452a56b0c3e75e7ca698273434e32945304

PR #526, opened 2023-02-09
Fix repeated calls to `Table.convert()`

### LINKED ISSUE #525: Repeated calls to `Table.convert()` fail
## Summary
When using the API, repeated calls to `Table.convert()`  do not work correctly since all conversions quietly use the callable (function, lambda) from the first call to `convert()` only. Subsequent invocations with different callables use the callable from the first invocation only.

## Example
```python

db = Database(memory=True)
table = db['table']
col = 'x'
table.insert_all([{col: 1}])
print(table.get(1))

table.convert(col, lambda x: x*2)
print(table.get(1))

def zeroize(x):
    return 0
#zeroize = lambda x: 0
#zeroize.__name__ = 'zeroize'
table.convert(col, zeroize)
print(table.get(1))
```

Output:
```
{'x': 1}
{'x': 2}
{'x': 4}
```
Expected:
```
{'x': 1}
{'x': 2}
{'x': 0}
```

## Explanation
This is some relevant [documentation](https://github.com/simonw/sqlite-utils/blob/1491b66dd7439dd87cd5cd4c4684f46eb3c5751b/docs/python-api.rst#registering-custom-sql-functions:~:text=By%20default%20registering%20a%20function%20with%20the%20same%20name%20and%20number%20of%20arguments%20will%20have%20no%20effect).

 * `Table.convert()` takes a `Callable` to perform data conversion on a column
 * The `Callable` is passed to `Database.register_function()`
 * `Database.register_function()` uses the callable's `__name__` attribute for registration
 * (Aside: all lambdas have a `__name__` of `<lambda>`: I thought this was the problem, and it was close, but not quite)
 * However `convert()` first wraps the callable by local function [`convert_value()`](https://github.com/simonw/sqlite-utils/blob/fc221f9b62ed8624b1d2098e564f525c84497969/sqlite_utils/db.py#L2661)
 * Consequently `register_function()` sees name `convert_value` for all invocations from `convert()`
 * `register_function()` silently ignores registrations using the same name, retaining only the first such registration

There's a mismatch between the comments and the code: https://github.com/simonw/sqlite-utils/blob/fc221f9b62ed8624b1d2098e564f525c84497969/sqlite_utils/db.py#L404

but actually the existing function is returned/used instead (as the "registering custom sql functions" doc I linked above says too). Seems like this can be rectified to match the comment?

## Suggested fix
I think there are four things:
1. The call to `register_function()` from `convert()`should have an explicit `name=` parameter (to continue using `convert_value()` and the progress bar).
2. For functions, this name can be the real function name. (I understand the sqlite api needs a name, and it's nice if those are recognizable names where possible). For lambdas would `'lambda-{uuid}'` or similar be acceptable? 
3. `register_function()` really should throw an error on repeated attempts to register a duplicate (function, arity)-pair.
4. A test? I haven't looked at the test framework here but seems this should be testable.

## See also 
- #458

mcarpenter:
Fixes #525. All tests pass.

There's perhaps a better way to name lambdas? There could be a collision if a caller passes a function with name like `lambda_123456`.

SQLite [documentation](https://www.sqlite.org/appfunc.html) is a little, ah, lite on function name specs. If there is a character that can be used in place of underscore in a SQLite function name that is not permitted in a Python function identifier then that could be a good way to prevent accidental collisions. (I tried dash, colon, dot, no joy).

Otherwise, there is little chance of this happening and if it should happen the risk is mitigated by now throwing an exception in the case of a (name, arity) collision without `replace=True`.

----
:books: Documentation preview :books:: https://sqlite-utils--526.org.readthedocs.build/en/526/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2664,13 +2664,16 @@ class Table(Queryable):
                     return v
                 return jsonify_if_needed(fn(v))
 
-            self.db.register_function(convert_value)
+            fn_name = fn.__name__
+            if fn_name == '<lambda>':
+                fn_name = f'lambda_{hash(fn)}'
+            self.db.register_function(convert_value, name=fn_name)
             sql = "update [{table}] set {sets}{where};".format(
                 table=self.name,
                 sets=", ".join(
                     [
-                        "[{output_column}] = convert_value([{column}])".format(
-                            output_column=output or column, column=column
+                        "[{output_column}] = {fn_name}([{column}])".format(
+                            output_column=output or column, column=column, fn_name=fn_name
                         )
                         for column in columns
                     ]
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -219,6 +219,11 @@ class AlterError(Exception):
     pass
 
 
+class FunctionAlreadyRegistered(Exception):
+    "A function with this name and arity was already registered"
+    pass
+
+
 class NoObviousTable(Exception):
     "Could not tell which table this operation refers to"
     pass
@@ -409,7 +414,7 @@ class Database:
             fn_name = name or fn.__name__
             arity = len(inspect.signature(fn).parameters)
             if not replace and (fn_name, arity) in self._registered_functions:
-                return fn
+                raise FunctionAlreadyRegistered(f'Already registered function with name "{fn_name}" and identical arity')
             kwargs = {}
             registered = False
             if deterministic:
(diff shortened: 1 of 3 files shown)


sqlite_utils/db.py
```diff
@@ -409,7 +414,7 @@ def register(fn):
             fn_name = name or fn.__name__
             arity = len(inspect.signature(fn).parameters)
             if not replace and (fn_name, arity) in self._registered_functions:
-                return fn
+                raise FunctionAlreadyRegistered(f'Already registered function with name "{fn_name}" and identical arity')
```
simonw, line 417: Just realized this represents a backward-incompatible change, so I'm going to back this out of the release.

PR #528, opened 2023-02-10
Enable `Table.convert()` on falsey values

### LINKED ISSUE #527: `Table.convert()` skips falsey values
# Summary

By design, `Table.convert()` does [not attempt](https://github.com/simonw/sqlite-utils/blob/fc221f9b62ed8624b1d2098e564f525c84497969/sqlite_utils/db.py#L2663) conversion of falsey values (`None`, `""`, `0`, ...). This is surprising (directly contradicts the docstring) and `convert()` may quietly skip cells where the user assumed a conversion would take place. 

# Example
Increment a column of integers by one

``` python

db = Database(memory=True)
table = db['table']
col = 'x'
table.insert_all([{col: 0}, {col:1}])
print(table.get(1)) # 0
print(table.get(2)) # 1
print()

table.convert(col, lambda x: x+1)
print(table.get(1)) # got 0, expected 1 ⚠⚠⚠
print(table.get(2)) # got 2, expected 2
```

Another example might be, say, transforming cells containing empty string to `NULL`.

# Discussion

This was, I think, a pragmatic choice so that consumers can skip writing guard clauses for these falsey values (particularly from the CLI). But this surprising undocumented behavior can lead to incorrect data. I don't think this is a good trade-off between convenience and correctness.

In the absence of this convenience users will either have to write guard clauses into their conversion expressions (or adapt the called function to do the same), so: 
``` python
    fn(value) if value else value
```
instead of:
``` python
    fn(value)
```
This is more typing and sometimes I will forget, and there will be errors. (But they will be noisy errors, which is a good thing).

Such a change will certainly inconvenience some existing consumers; there will be some breakage. But I think this is worth it to avoid quietly not converting some values by default, which can lead to quietly bad data.

I have a PR that I will attach, please take a look and see what you think.

mcarpenter:
Fixes #527

----
:books: Documentation preview :books:: https://sqlite-utils--528.org.readthedocs.build/en/528/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2660,8 +2660,6 @@ class Table(Queryable):
 
             def convert_value(v):
                 bar.update(1)
-                if not v:
-                    return v
                 return jsonify_if_needed(fn(v))
 
             self.db.register_function(convert_value)
--- a/sqlite_utils/recipes.py
+++ b/sqlite_utils/recipes.py
@@ -14,6 +14,8 @@ def parsedate(value, dayfirst=False, yearfirst=False, errors=None):
     - errors=r.IGNORE to ignore values that cannot be parsed
     - errors=r.SET_NULL to set values that cannot be parsed to null
     """
+    if not value:
+        return value
     try:
         return (
             parser.parse(value, dayfirst=dayfirst, yearfirst=yearfirst)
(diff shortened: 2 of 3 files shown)


simonw:
I fixed this in:
- #527

Will fully remove this misfeature in:
- #542

PR #531, opened 2023-03-11
Add paths for homebrew on Apple silicon

eyeseast:
This also passes in the extension path when specified in GIS methods. Wherever we know an extension path, we use `db.init_spatialite(find_spatialite() or load_extension)`.

----
:books: Documentation preview :books:: https://sqlite-utils--531.org.readthedocs.build/en/531/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1407,7 +1407,7 @@ def create_database(path, enable_wal, init_spatialite, load_extension):
 
     # load spatialite from expected locations and initialize metadata
     if init_spatialite:
-        db.init_spatialite()
+        db.init_spatialite(find_spatialite() or load_extension)
 
     db.vacuum()
 
@@ -2902,7 +2902,7 @@ def add_geometry_column(
     # load spatialite, one way or another
     if load_extension:
         _load_extensions(db, load_extension)
-    db.init_spatialite()
+    db.init_spatialite(find_spatialite() or load_extension)
 
     if db[table].add_geometry_column(
         column_name, geometry_type, srid, coord_dimension, not_null
@@ -2934,7 +2934,7 @@ def create_spatial_index(db_path, table, column_name, load_extension):
     # load spatialite
     if load_extension:
         _load_extensions(db, load_extension)
-    db.init_spatialite()
+    db.init_spatialite(find_spatialite() or load_extension)
 
     if column_name not in db[table].columns_dict:
         raise click.ClickException(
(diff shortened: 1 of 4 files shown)


sqlite_utils/cli.py
```diff
@@ -1407,7 +1407,7 @@ def create_database(path, enable_wal, init_spatialite, load_extension):
 
     # load spatialite from expected locations and initialize metadata
     if init_spatialite:
-        db.init_spatialite()
+        db.init_spatialite(find_spatialite() or load_extension)
```
simonw, line 1410: I don't understand this pattern.

The `init_spatialite()` function already calls `find_spatialite()` if necessary:

https://github.com/simonw/sqlite-utils/blob/fc221f9b62ed8624b1d2098e564f525c84497969/sqlite_utils/db.py#L1161-L1165

The `load_extension` option is a bit different: It's actually a list of extensions passed using `--load-extension path` on the command-line, which is used by this code:

https://github.com/simonw/sqlite-utils/blob/fc221f9b62ed8624b1d2098e564f525c84497969/sqlite_utils/cli.py#L3037-L3047

simonw:
Aah, I think I see why you wrote it like that.

The problem is that `init_spatialite()` does other stuff too:

https://github.com/simonw/sqlite-utils/blob/fc221f9b62ed8624b1d2098e564f525c84497969/sqlite_utils/db.py#L1161-L1171

So it needs to be able to load the SpatiaLite extension from the correct place, and THEN run `select InitSpatialMetadata()` to configure the database, if needed.

So the problem you're trying to solve here is to let people optionally pass in the path to SpatiaLite if it's not one of the ones that are searched by default.

eyeseast:
Exactly, that's what I was running into. On my M2 MacBook, SpatiaLite ends up in what is -- for the moment -- a non-standard location, so even when I passed in the location with `--load-extension`, I still hit an error on `create-spatial-index`.

What I learned doing this originally is that SQLite needs to load the extension for each connection, even if all the SpatiaLite stuff is already in the database. So that's why `init_spatialite()` gets called again.

Here's the code where I hit the error: https://github.com/eyeseast/boston-parcels/blob/main/Makefile#L30 It works using this branch.

I'm not attached to this solution if you can think of something better. And I'm not sure, TBH, my test would actually catch what I'm after here.

eyeseast:
I'm going to close this in favor of #536. Will try a cleaner approach to custom paths once that one is merge.

PR #537, opened 2023-04-12
Support self-referencing FKs in `Table.create`

numist:
----
:books: Documentation preview :books:: https://sqlite-utils--537.org.readthedocs.build/en/537/


simonw:
Thanks!

simonw:
Here's an example from the new tests:

https://github.com/numist/sqlite-utils/blob/a75abeb61b91a28650d3b9933e7ec80ad0d92529/tests/test_create.py#L291-L307

Issue #433, opened 2022-05-17
CLI eats my cursor


chapmanjacobd:
I'm not sure why this happens but `sqlite-utils` makes my terminal cursor disappear after running commands like `sqlite-utils insert`. I've only noticed this behavior in `sqlite-utils`. Not sure if it is a bug in `kitty`, `fish` or `sqlite-utils`

frafra:
I can confirm. This only happens with sqlite-utils. I am using gnome-terminal with bash.

simonw:
Marking this as help wanted because I can't figure out how to replicate it!

chapmanjacobd:
When I run `reset` it fixes my terminal. I suspect it is related to the progress bar

alecstein:
Came here to say that I also have this issue.

mcarpenter:
I think I see what is happening here, although I haven't quite work out a fix yet. Usually:

* `click.progressbar.render_progress()` renders the cursor invisible on each invocation (update of the bar)
* When the progress bar goes out of scope, the `__exit()__` method is invoked, which calls `render_finish()` to make the cursor re-appear.

(See terminal escape sequences `BEFORE_BAR` and `AFTER_BAR` in click).

However the sqlite-utils `utils.file_progress` context manager wraps `click.progressbar` and yields an instance of a helper class:

``` python
@contextlib.contextmanager     
def file_progress(file, silent=False, **kwargs):
   ...
        with click.progressbar(length=file_length, **kwargs) as bar:
            yield UpdateWrapper(file, bar.update) 
```

The yielded `UpdateWrapper` goes out of scope quickly and `click.progressbar.__exit__()` is called. The cursor is made un-invisible. Hoewever `bar` is still live and so when the caller iterates on the yielded wrapper this invokes the bar's update method, calling `render_progress()`, each time printing the "make cursor invisible" escape code. The `progressbar.__exit__` is not called again, so the cursor doesn't re-appear.

jonafato:
I also ran into this recently. See below for a patch for one possible solution (tested via "it works on my machine", but I don't expect that this behavior would vary a whole lot across terminal emulators and shells). Another possible solution might be to subclass click's `ProgressBar` to keep the logic within the original context manager. Happy to send a PR or for this patch to serve as the basis for a fix that someone else authors.

```patch
diff --git a/sqlite_utils/utils.py b/sqlite_utils/utils.py
index 06c1a4c..530a3a3 100644
[...]

PR #598, opened 2023-10-04
Fixed issue #433 - CLI eats cursor

spookylukey:
The issue is that underlying iterator is not fully consumed within the body of the `with file_progress()` block. Instead, that block creates generator expressions like `docs = (dict(zip(headers, row)) for row in reader)`

These iterables are consumed later, outside the `with file_progress()` block, which consumes the underlying iterator, and in turn updates the progress bar.

This means that the `ProgressBar.__exit__` method gets called before the last time the `ProgressBar.update` method gets called. The result is that the code to make the cursor invisible (inside the `update()` method) is called after the cleanup code to make it visible (in the `__exit__` method).

The fix is to move consumption of the `docs` iterators within the progress bar block. (

(An additional fix, to make ProgressBar more robust against this kind of misuse, would to make it refusing to update after its `__exit__` method had been called, just like files cannot be `read()` after they are closed. That requires a in the click library).

Note that Github diff obscures the simplicity of this diff, it's just indenting a block of code.

----
:books: Documentation preview :books:: https://sqlite-utils--598.org.readthedocs.build/en/598/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1024,93 +1024,97 @@ def insert_upsert_implementation(
             if flatten:
                 docs = (_flatten(doc) for doc in docs)
 
-    if stop_after:
-        docs = itertools.islice(docs, stop_after)
+        if stop_after:
+            docs = itertools.islice(docs, stop_after)
 
-    if convert:
-        variable = "row"
-        if lines:
-            variable = "line"
-        elif text:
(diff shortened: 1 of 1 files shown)


simonw:
Thanks!

simonw:
Manually tested. Before:

cursor-bug

After:

cursor-fix

PR #609, opened 2024-01-02
Enable FTS for attached dbs

### LINKED ISSUE #608: Attached database tables representable by `Table`
`attach`ing is one of the first things documented, but both from playing around a bit and looking at the code, it seems like while attaching affects the underlying connection, there's no way to construct a Table object from an attached db+table name.

Because of this, the rest of the documentation doesn't appear to apply.

Concretely, my use case is that I'd like to be able to attach a read-only database but use `sqlite-utils` for the "easy mode" full text search that it builds in. What that would mean is creating my own database full of indexed search tables based on the data from attached tables, and then using the same rowid-based joins that `sqlite-utils` performs under the hood when actually calling `table.search()`.

From my code review so far, it seems like in order to do this we would need to add an optional `schema` parameter to the Table objects, as well as a list of attached schema names, such that it would be possible to do the following:

```
db = Database('foo.db')  # contains table `foot`
db.attach('bar', 'bar.db')  # contains table `bart`

bart = db.table('bar.bart')  # auto-parse the schema.table_name syntax the same way SQLite does?
# bart would support all operations supported on a table in the 'main' database.
# e.g. 
bart.enable_fts(['name', 'age'])  # would create the necessary tables - but inside the main database, since otherwise it would make more sense to have a separate connection to `bar` and perform the operation directly over there.
```

I could imagine that there might be some sharp edges here, since there are a few cross-database limitations in SQLite, and certain operations that Tables support might not make sense if they could be referenced in this way. For instance, the trigger operations for full-text-search (which don't apply to my use case but are documented capabilities of sqlite-utils) would be a little weird, because I believe they would have to be temp triggers which would expire as soon as the database was detached. It might also be a little tricky to decide exactly how to name the in-database FTS searches so that a separate process provided with both databases would be able to recognize that sqlite-utils had already created the necessary FTS tables - but this doesn't seem insurmountable to me.

If the maintainers think this would be feasible but don't have time to dedicate to building in the support, I'd be willing to get started on a PR.

petergaultney:
This is just a work-in-progress, although it does:

- Allows tables in attached databases to be 'addressed' by `sqlite-utils` by loading them in explicitly: `db.attach('./a1.db', alias='a1'); a1_foo = db.table('foo', schema='a1')`. This table should generally operate successfully, although this is far from fully tested and I expect there are some edge cases where it does not work as expected due to SQLite limitations on attached tables.
- Permits creation of and basic usage of FTS on table(s) in an attached db, where the FTS tables are created in the `main` database.
  - Continues to store all the true state inside the database itself - e.g., detects existing FTS tables and automagically allows searching if the database is detached and later reattached.

All tests pass locally, as well as black and mypy.

You could call this a proof-of-concept; I think there probably would need to be discussions about some of my refactors around triggers, counts, etc, to decide whether we want to take a similar approach to those (modify the 'main' database rather than the attached ones, under the reasoning that someone attaching a database could presumably make separate connections to that database if they wanted to set up triggers, count tables, etc), do something else, or possibly try to just not support anything having to do with attached databases/tables in those cases.

This would close #608 

<hr />

2024-01-04: Only now am I realizing, based on the content of your tests, that you have use cases where you want to support dots being part of an actual table name. And that this means my approach is not workable, since the API would not allow us to distinguish between cases where the user intended to match an existing, attached database name, and cases where the user wanted to put that matching string in a table name inside `main`.

Hopefully the idea of supporting this is compelling enough that we can figure out what a better API (and internal storage format) might be. Perhaps we could use tuples `("a1", "foo")` and force users to be explicit when creating/accessing the tables from the `Database` object, and then store the database name in a separate field inside the `Queryable` object?

I really liked the simplicity of a simple string (it works well for the CLI use cases as well, although maybe a CLI would never attach a database in the first place, since that's a stateful operation)  -- but I am sure you don't want to break backward compatibilty.

<hr />

2024-01-09: I have reworked this so that the above section is no longer an issue. The `schema_name` is stored explicitly within the Table object, and the only real change to functionality is that `table_names()` and related methods on `Database` will no longer operate across schemas - they instead operate on a single schema (attached database) at a time.

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -532,7 +532,9 @@ class Database:
             self._tracer(sql, None)
         return self.conn.executescript(sql)
 
-    def table(self, table_name: str, **kwargs) -> Union["Table", "View"]:
+    def table(
+        self, table_name: str, schema: str = "", **kwargs
+    ) -> Union["Table", "View"]:
         """
         Return a table object, optionally configured with default options.
 
@@ -541,10 +543,10 @@ class Database:
         :param table_name: Name of the table
         """
         if table_name in self.view_names():
-            return View(self, table_name, **kwargs)
+            return View(self, table_name, schema_name=schema, **kwargs)
         else:
             kwargs.setdefault("strict", self.strict)
-            return Table(self, table_name, **kwargs)
+            return Table(self, table_name, schema_name=schema, **kwargs)
 
     def quote(self, value: str) -> str:
         """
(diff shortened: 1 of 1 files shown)


sqlite_utils/db.py
```diff
@@ -1271,12 +1288,34 @@ def init_spatialite(self, path: Optional[str] = None) -> bool:
         return result and bool(result[0])
 
 
+def _split_names(fullname: str) -> Tuple[str, str]:
+    if '.' not in fullname:
+        return '', fullname
```
petergaultney, line 1293: this is a choice to not make all the existing SQL more verbose by putting `main` everywhere - not only does this keep the SQL 'quieter', but it avoids having to refactor a tooon of tests that are asserting string values of SQL queries.

sqlite_utils/db.py
```diff
@@ -1292,7 +1331,7 @@ def count_where(
         :param where_args: Parameters to use with that fragment - an iterable for ``id > ?``
           parameters, or a dictionary for ``id > :id``
         """
-        sql = "select count(*) from [{}]".format(self.name)
+        sql = "select count(*) from {}".format(escaped_name(self.name))
```
petergaultney, line 1334: you'll see this pattern lots of places. Basically, for attached tables, you cannot use the `[table_name]` escape syntax if the table name is actually `attached_db.table1` - it has to be `attached_db.[table1]`.

sqlite_utils/db.py
```diff
@@ -1387,12 +1426,23 @@ def pks_and_rows_where(
...
+    @property
+    def is_attached(self) -> bool:
+        return dbname(self.name) not in {'', 'main'}
+
+    @property
+    def _pragma_name(self) -> Tuple[str, str]:
+        if "." in self.name:
+            db, name = self.name.split(".")
+            return db + ".", name
+        return "", self.name
```
petergaultney, line 1438: this is maybe premature optimization for not changing too much of the code in an visually noisy way. It really just lets me modify all the places we do PRAGMA things without making big changes to what's there.

sqlite_utils/db.py
```diff
@@ -1709,9 +1763,9 @@ def duplicate(self, new_name: str) -> "Table":
         if not self.exists():
             raise NoTable(f"Table {self.name} does not exist")
         with self.db.conn:
-            sql = "CREATE TABLE [{new_table}] AS SELECT * FROM [{table}];".format(
```
petergaultney, line 1712: lots of these refactors to just try to generally allow the escaping to work as intended while still allowing `table.name` to include a single period. Which, incidentally, I am not validating, and probably I should be validating "at most one ." as well as "if there's a dot, the database name matches something that's actually in the current connection" - although `sqlite-utils`'s lazy-ish approach to verifying state when creating Table objects makes me a little uncertain where we'd want to apply that validation.

sqlite_utils/db.py
```diff
@@ -2386,6 +2444,7 @@ def enable_fts(
         :param tokenize: Custom SQLite tokenizer to use, for example ``"porter"`` to enable Porter stemming.
         :param replace: Should any existing FTS index for this table be replaced by the new one?
         """
+        table_name = tablename(self.name)
```
petergaultney, line 2447: in the end, the actual changes to enable FTS were, as expected, essentially nil outside the changes to support 'addressing' attached Databases and their Tables.

Hopefully this new ability can be an overall win for the library, and hopefully since you have a lot of experience with SQLite you can help advise on some of the pitfalls I am missing.

sqlite_utils/db.py
petergaultney: my comment about validation is no longer relevant, since I'm storing the schema name explicitly.

PR #618, opened 2024-03-04
Use context managers for file writes in tests

kbaikov:
This removes the need to explicitly close files
and avoids accidental unclosed files

----
📚 Documentation preview 📚: https://sqlite-utils--618.org.readthedocs.build/en/618/


simonw:
Thanks, this is neater.

Issue #541, opened 2023-05-08
Get tests to pass with `pytest -Werror`


simonw:
Inspired by:
- #534

simonw:
There are 8 failing tests left:

```
==== short test summary info ====
FAILED tests/test_cli_memory.py::test_memory_csv[False-test] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
FAILED tests/test_cli_memory.py::test_memory_csv[False-t] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
FAILED tests/test_cli_memory.py::test_memory_csv[False-t1] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
FAILED tests/test_cli_memory.py::test_memory_tsv[False] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
FAILED tests/test_cli_memory.py::test_memory_dump[extra_args0] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
FAILED tests/test_cli_memory.py::test_memory_two_files_with_same_stem - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
FAILED tests/test_recipes.py::test_dateparse_errors[None-parsedate] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function Table.convert.<locals>.convert_value at 0x106bcaca0>
FAILED tests/test_recipes.py::test_dateparse_errors[None-parsedatetime] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function Table.convert.<locals>.convert_value at 0x106bc9620>
ERROR tests/test_cli.py::test_upsert_analyze - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
==== 8 failed, 894 passed, 4 skipped, 1 error in 6.27s ====
```
Full traceback here: https://gist.github.com/simonw/b40b3e814729d6c02a0302a84ce54d9e

lbellomo:
I made a new run and found a few more cases. I can try to make a PR to fix them.

```
==== short test summary info ====
FAILED tests/test_cli.py::test_upsert_analyze - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO ...
[...]

PR #623, opened 2024-03-31
Fix test to pass `pytest -Werror`

lbellomo:
Refs https://github.com/simonw/sqlite-utils/issues/541

Still TODO:
- [x] Fix tests/test_cli.py::test_upsert_analyze
- [x] Fix tests/test_cli_memory.py::test_memory_csv
- [x] Fix tests/test_cli_memory.py::test_memory_tsv
- [x] Fix tests/test_cli_memory.py::test_memory_dump
- [x] Fix tests/test_cli_memory.py::test_memory_two_files_with_same_stem
- [x] Fix tests/test_insert_files.py::test_insert_files
- [ ] Fix tests/test_recipes.py::test_dateparse_errors

----
📚 Documentation preview 📚: https://sqlite-utils--623.org.readthedocs.build/en/623/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -865,7 +865,7 @@ def insert_upsert_options(*, require_pk=False):
                     required=True,
                 ),
                 click.argument("table"),
-                click.argument("file", type=click.File("rb"), required=True),
+                click.argument("file", type=click.File("rb", lazy=True), required=True),
                 click.option(
                     "--pk",
                     help="Columns to use as the primary key, e.g. id",
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1,7 +1,7 @@
 import base64
 import click
 from click_default_group import DefaultGroup  # type: ignore
-from datetime import datetime
+from datetime import datetime, timezone
 import hashlib
 import pathlib
 from runpy import run_module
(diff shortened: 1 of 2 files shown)


lbellomo:
The first case is not the test you mark, but the previous [test_upsert_pk_required](https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/tests/test_cli.py#L1058). When you run the upsert "click" opens the file first but as it then gives the error `"Error: Missing option '--pk"` it doesn't close it correctly.

https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/sqlite_utils/cli.py#L868

 The solution is to pass `lazy=True` to the [click.File](https://click.palletsprojects.com/en/8.1.x/api/#click.File) so that it does not open it until it is necessary.

lbellomo:
It seems that in [test_memory_csv](https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/tests/test_cli_memory.py#L17) only for when `use_stdin=False` the `io.TextIOWrapper` on line 297 is not closed correctly.

https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/sqlite_utils/utils.py#L295-L302

The only way I can think of is to pass the `buffer to _extra_key_strategy` and close it once the generator is finished, but surely there must be a nicer way.

lbellomo:
Similar to the first case, it is not `test_memory_dump` but [`test_memory_csv_encoding`](https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/tests/test_cli_memory.py#L119) the previous test that is the culprit, and also only when `use_stdin=False`.

It seems that when the enconding is wrong it raises a `UnicodeDecodeError` error when it tries to iterate over the generator from [`rows_from_file`](https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/sqlite_utils/utils.py#L236) and that's why it doesn't reach the end to close the `fp` file.

https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/sqlite_utils/cli.py#L1949-L1968

The only thing that I can think of is to catch the error, close the file and raise it again.

lbellomo:
[test_insert_files.py](https://github.com/simonw/sqlite-utils/blob/1d050dcdc7c8099108abd2987932a3d8d7bf55aa/tests/test_insert_files.py#L17) is a deprecation warning. `datetime.utcfromtimestamp` is replaced by [`datetime.fromtimestamp`](https://docs.python.org/3.12/library/datetime.html#datetime.datetime.fromtimestamp) in py3.12, I pass the timezone to make sure it is utc (otherwise it grabs the local one) and stripe the timezone to be the same as it was before.

Issue #159, opened 2020-09-16
delete_where does not commit...


spdkils:
When you use the delete_where() function on a table, it never commits....

Is that intentional?

simonw:
See #121 - I need to think harder about how this all interacts with transactions.

You can do this:

```python
with db.conn:
    db["mytable"].delete_where()
```
But that should be documented and maybe rethought.

spdkils:
I appreciate the response, it's just unexpected.

If I insert, it commits, if I update it commits, if I upsert it commits... if I delete.. it doesn't???
Confused me...

I did just db commit it... But it's confusing.

simonw:
Yeah I'm going to class this as a bug - that's definitely confusing.

simonw:
Independent of the transaction changes in #121 I may be able to check `self.conn.in_transaction` to see if a transaction is active and, if one is NOT active, execute the delete inside of one. https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.in_transaction

simonw:
That's strange... this test here doesn't manually commit a transaction and passes:

https://github.com/simonw/sqlite-utils/blob/7805d53bcf11199bd1f2b07e05ae90151f9d0eb0/tests/test_delete.py#L17-L23

simonw:
@spdkils can you share a minimal code example that exhibits the behavior you're seeing?

limar:
I confirm the bug. Happens for me in version 3.6. I use the call to delete all the records:
`table.delete_where()`
This does not delete anything. 

I see that `delete()` method DOES use context manager `with self.db.conn:` which should help. You may want to align the code of both methods.

dracos:
Just stumbled across this, wondering why none of my deletes were working.

dracos:
In a related issue, nearly a year later I just stumbled across this again, as I wondered why none of my rebuild-fts were rebuilding. It looks like: disable_fts in db.py commits; enable_fts partly commits except the last step (due to executescript committing a pending transaction); rebuild_fts won't commit unless manually done as above with e.g. a context manager.

simonw:
Fixed this in https://github.com/simonw/sqlite-utils/commit/b17c37144eded6af1a326699e8fa732b039ff4be

PR #630, opened 2024-07-03
Documented the use of `delete_where`

learning4life:
Documented the use of delete_where, as shown in https://github.com/simonw/sqlite-utils/issues/159

----
📚 Documentation preview 📚: https://sqlite-utils--630.org.readthedocs.build/en/630/


simonw:
Thanks, I'll land this change but I'd like to fix this at source later on.

Issue #633, opened 2024-07-21
Table indexes are silently dropped from tables when foreign keys are added


matdmiller:
After creating a table with an index, if you add a foreign key constraint to the table it drops existing non-related indexes. 

I have attached a minimal repro example screenshot and code.

<img width="1069" alt="Screenshot 2024-07-20 at 8 05 26 PM" src="https://github.com/user-attachments/assets/e5d6cbe9-0c26-4430-aa92-75b85b2cafea">

Repro code
```python
from sqlite_utils import Database

from pathlib import Path
db_fp = './db_repro_index.db'
Path(db_fp).unlink(missing_ok=True), Path(f'{db_fp}-shm').unlink(missing_ok=True),Path(f'{db_fp}-wal').unlink(missing_ok=True);

# db = database(db_fp)
db = Database(db_fp)

db['foo'].create(dict(id=int, name=str), pk='id', not_null=['name'])

db['foo'].create_index(['name'], unique=True)
db['foo'].indexes

db['bar'].create(dict(id=int, name=str, foo_id=int), pk='id', not_null=['name'])

db['bar'].create_index(['name'], unique=True)
db['bar'].indexes

db.add_foreign_keys((('bar','foo_id','foo','id'),))

db['bar'].indexes
```

matdmiller:
Looks like when adding fk constraints the table is dropped and recreated, but the indexes are not. The tracer functionality is quite helpful for debugging! 

<img width="486" alt="Screenshot 2024-07-20 at 9 07 44 PM" src="https://github.com/user-attachments/assets/672d66bd-49c3-48d4-a746-2dd6ec508145">
<img width="679" alt="Screenshot 2024-07-20 at 9 07 18 PM" src="https://github.com/user-attachments/assets/ac52790b-6027-46a4-8df6-31005c7f0eaa">

matdmiller:
I think it is best to try and preserve existing indexes and fail loudly if they are unable to be recreated automatically. At the very least I believe you should be warned if they are not going to be recreated.

Below is my proposed solution with code:

1. If a table is kept it will automatically drop the original index. To be able to create the index on the new table using the original `CREATE INDEX` statement the original index must be dropped. I think it's a reasonable default for the 'kept' table to lose its indexes. I suspect the most common reason to keep the original table is for 'backup' purposes. It would also be difficult to automatically create updated `CREATE INDEX` statements for the `kept` renamed table where the statements are more complicated.
[...]

PR #634, opened 2024-07-21
Recreate indexes when calling transform when possible and raise an er…

matdmiller:
Recreate indexes when calling transform when possible and raise an error when they cannot be retained automatically.

This resolves my initial problem: 'After creating a table with an index, if you add a foreign key constraint to the table it drops existing non-related indexes.'

GH Issue:
- https://github.com/simonw/sqlite-utils/issues/633

My explanation, thoughts and reasoning:

I think it is best to try and preserve existing indexes and fail loudly if they are unable to be recreated automatically. At the very least I believe you should be warned if they are not going to be recreated.

Below is my proposed solution with code:

1) If a table is kept it will automatically drop the original index. To be able to create the index on the new table using the original CREATE INDEX statement the original index must be dropped. I think it's a reasonable default for the 'kept' table to lose its indexes. I suspect the most common reason to keep the original table is for 'backup' purposes. It would also be difficult to automatically create updated CREATE INDEX statements for the kept renamed table where the statements are more complicated.

2) If a column exists in the original index but not in the new table, you should have to manually recreate the index as desired. In complicated CREATE INDEX statements it is difficult to automatically replace or remove columns correctly. Failing before any operations have been completed is desired because it allows you to know there will be a problem prior to the SQL statements being executed allowing you to collect the required information required to manually update the indexes prior to any actions taking place as well as not leaving your database in a potentially invalid state.

----
📚 Documentation preview 📚: https://sqlite-utils--634.org.readthedocs.build/en/634/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -156,6 +156,10 @@ XIndexColumn = namedtuple(
 Trigger = namedtuple("Trigger", ("name", "table", "sql"))
 
 
+class TransformError(Exception):
+    pass
+
+
 ForeignKeyIndicator = Union[
     str,
     ForeignKey,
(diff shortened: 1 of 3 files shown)


sqlite_utils/db.py
```diff
@@ -1967,6 +1967,28 @@ def transform_sql(
...
             "ALTER TABLE [{}] RENAME TO [{}];".format(new_table_name, self.name)
         )
+        # Re-add existing indexes
+        for index in self.indexes:
+            if index.origin not in ("pk"):
+                index_sql = self.db.execute(
+                    """SELECT sql FROM sqlite_master WHERE type = 'index' AND name = :index_name;""",
+                    {"index_name": index.name},
+                ).fetchall()[0][0]
+                assert index_sql is not None, (
```
simonw, line 1977: The problem with `assert` for this kind of thing is that these won't execute if Python is run in optimize mode: https://docs.python.org/3/using/cmdline.html#cmdoption-O

I don't personally like that feature, but I think we should switch to a different exception just in case.

sqlite_utils/db.py
```diff
@@ -1967,6 +1967,28 @@ def transform_sql(
         sqls.append(
             "ALTER TABLE [{}] RENAME TO [{}];".format(new_table_name, self.name)
         )
+        # Re-add existing indexes
+        for index in self.indexes:
+            if index.origin not in ("pk"):
```
simonw, line 1972: I changed this to 'index.origin != "pk"` instead.

simonw:
I tried running this through OpenAI's `o1-preview` as an experiment, specifically asking if the tests covered all of the cases: https://gist.github.com/simonw/edcd047a703ab6d6759a5a207364d744#response

It cost me $1.50 for the two prompts! Still reviewing the o1 produced test code to see if I think it's worth including.

simonw:
@matdmiller thanks very much for this - I'll have it out in a release shortly.

Issue #699, opened 2025-12-21
Make square braces only return tables, not views


simonw:
This syntactic sugar:

    db["table_name"].insert(...)

Had caused all sorts of type-related mess as it can return either a table or a view, and views can't do insert etc.

For 4.0 I can break backwards compatibility, and making it so views can no longer be returned by square brace access is a small enough breaking change that makes all of the type stuff much nicer and hence is worthwhile I think.

Kyli33:
Hi

simonw:
I've decided not to do this. I'm going to have `db[name]` continue to return either tables OR views, but if you know you want a table or view specifically (in a typed project) you can use `db.table(name)` or `db.view(name)` instead.

PR #700, opened 2025-12-21
Make square braces only return tables, not views

simonw:
Breaking change for 4.0: db["name"] now only returns Table objects, never View objects.
This improves type safety since views lack methods like .insert().

Use db.view("view_name") to access views explicitly.

Closes:
- #699

----
📚 Documentation preview 📚: https://sqlite-utils--700.org.readthedocs.build/en/700/

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -41,3 +41,4 @@ Contents
    cli-reference
    contributing
    changelog
+   upgrading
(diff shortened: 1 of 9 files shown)


simonw:
I decided to go the other way, square braces still return both, but documentation encourages `db.table()` instead.

PR #707, opened 2026-02-07
Fix: Handle CSV with only header row in --detect-types (fixes #702)

### LINKED ISSUE #702: Insert CSV with --detect-types crashes for CSV files that contain only a header row
I run into this error:
    insert_upsert_implementation(

RamiNoodle733:
This PR fixes issue #702 where inserting a CSV file with only a header row (no data rows) would crash when using --detect-types.

## Problem
When a CSV contains only headers and no data rows:
1. The TypeTracker wraps the docs iterator but no rows are tracked
2. After insert_all, the table doesn't exist (no rows were inserted)
3. Calling transform(types=tracker.types) fails because the table doesn't exist

## Solution
Added a check to ensure the table exists before attempting to apply type transformations:
- Changed `if tracker is not None:` to `if tracker is not None and db.table(table).exists():`
- Applied in both `insert_upsert_implementation()` and `memory()` commands

## Changes
- Modified `sqlite_utils/cli.py`: Added existence check before transform
- Added test `test_insert_csv_headers_only` to verify the fix

## Testing
- Added new test case that verifies CSV with only headers doesn't crash
- Verified the fix resolves the exact error from issue #702

----
📚 Documentation preview 📚: https://sqlite-utils--707.org.readthedocs.build/en/707/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1176,7 +1176,7 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        if tracker is not None and db.table(table).exists():
             db.table(table).transform(types=tracker.types)
 
         # Clean up open file-like objects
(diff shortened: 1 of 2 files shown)


simonw:
Thanks for this.

PR #708, opened 2026-02-07
Fix #705: Clarify ensure_autocommit_off() naming and documentation

### LINKED ISSUE #705: ensure_autocommit_off() is confusing
[ensure_autocommit_off()](https://github.com/simonw/sqlite-utils/blob/e66e82f2c8b3835a316de6ed1407f1a345358e41/sqlite_utils/db.py#L390) is somewhat confusing:

The docstring states:
> Ensure autocommit is off for this database connection.

However, the codes sets `self.conn.isolation_level = None`, which is documented in the Python stdlib as:
> If [isolation_level](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.isolation_level) is set to None, no transactions are implicitly opened at all. This leaves the underlying SQLite library in [autocommit mode](https://www.sqlite.org/lang_transaction.html#implicit_versus_explicit_transactions),

https://docs.python.org/3/library/sqlite3.html#sqlite3-transaction-control-isolation-level

So that means that `ensure_autocommit_off()` actually turns on the sqlite3 in autocommit mode.

AFAIU, in order for `ensure_autocommit_off()` to actually disable autocommit, it would also need to issue a `BEGIN` command before `yield`, so that an explicit transaction is started:
> If [isolation_level](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.isolation_level) is set to None, no transactions are implicitly opened at all. This leaves the underlying SQLite library in [autocommit mode](https://www.sqlite.org/lang_transaction.html#implicit_versus_explicit_transactions), but also allows the user to perform their own transaction handling using explicit SQL statements.

https://docs.python.org/3/library/sqlite3.html#sqlite3-transaction-control-isolation-level

The last part `but also allows the user to perform their own transaction handling using explicit SQL statements.` is what an explicit `BEGIN` would allow.

RamiNoodle733:
Fixes #705

The function name `ensure_autocommit_off()` was confusing because it sets `isolation_level = None`, which actually ENABLES autocommit mode in SQLite, not disables it.

## Changes

- **Added** new `ensure_autocommit_on()` method with correct name and accurate docstring explaining that `isolation_level = None` enables autocommit mode
- **Deprecated** `ensure_autocommit_off()` - kept as alias for backwards compatibility with `DeprecationWarning`
- **Updated** internal usages (`enable_wal()`, `disable_wal()`) to use the new method name
- **Added** `warnings` import for the deprecation notice

## Backwards Compatibility

Existing code using `ensure_autocommit_off()` will continue to work but will emit a `DeprecationWarning` directing users to the new method name.

----
📚 Documentation preview 📚: https://sqlite-utils--708.org.readthedocs.build/en/708/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1176,7 +1176,7 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        if tracker is not None and db.table(table).exists():
             db.table(table).transform(types=tracker.types)
 
         # Clean up open file-like objects
@@ -2033,7 +2033,7 @@ def memory(
                 rows = (_flatten(row) for row in rows)
 
             db.table(file_table).insert_all(rows, alter=True)
-            if tracker is not None:
+            if tracker is not None and db.table(file_table).exists():
                 db.table(file_table).transform(types=tracker.types)
             # Add convenient t / t1 / t2 views
             view_names = ["t{}".format(i + 1)]
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -40,6 +40,7 @@ from typing import (
     Tuple,
 )
 import uuid
+import warnings
 from sqlite_utils.plugins import pm
 
 try:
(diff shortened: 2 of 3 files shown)


simonw:
Superseded by `ensure_autocommit_on()` implementation shipped in 4.0.

PR #709, opened 2026-02-07
Fix enable_fts() when FTS table already exists without replace=True

### LINKED ISSUE #694: sqlite-utils 4.0a1 enable_fts(replace=True): sqlite3.OperationalError: table "{}_fts" already exists
I've been playing around with the sqlite-utils v4 beta and everything seems to mostly work but there seems to be different behavior with `enable_fts()` compared to v3.39.

```py
### https://github.com/chapmanjacobd/library/blob/849349db7c27aae54afde234cf3e734fb41c85e5/library/utils/db_utils.py#L217
                db[table].enable_fts(
                    fts_columns,
                    create_triggers=True,
                    replace=True,
                    tokenize=(
                        "trigram" if sqlite3.sqlite_version_info >= (3, 34, 0) else 'unicode61 "tokenchars=_."'
                    ),  # https://www.sqlite.org/releaselog/3_34_0.html
                )
```

This code runs fine on 3.39 but in 4.0a1 I get an error like this:

```py
  File "/home/xk/.local/lib/python3.14/site-packages/sqlite_utils/db.py", line 2567, in enable_fts
    self.db.executescript(create_fts_sql)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/xk/.local/lib/python3.14/site-packages/sqlite_utils/db.py", line 561, in executescript
    return self.conn.executescript(sql)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^
sqlite3.OperationalError: table "playlists_fts" already exists
```

RamiNoodle733:
When calling `enable_fts()` on a table that already has an FTS index, without passing `replace=True`, the function now returns early instead of attempting to create the table again and raising an `sqlite3.OperationalError: table '{}_fts' already exists` error.

## Changes
- Added check in `enable_fts()` to return early if FTS table exists and `replace=False` (the default)
- Added regression test `test_enable_fts_twice_without_replace()`

Fixes #694

----
📚 Documentation preview 📚: https://sqlite-utils--709.org.readthedocs.build/en/709/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2572,7 +2572,11 @@ class Table(Queryable):
             )
         )
         should_recreate = False
-        if replace and self.db["{}_fts".format(self.name)].exists():
+        fts_table_exists = self.db["{}_fts".format(self.name)].exists()
+        if not replace and fts_table_exists:
+            # FTS table already exists and replace=False, so return early
+            return self
+        if replace and fts_table_exists:
             # Does the table need to be recreated?
             fts_schema = self.db["{}_fts".format(self.name)].schema
             if fts_schema != create_fts_sql:
--- a/tests/test_fts.py
+++ b/tests/test_fts.py
@@ -428,6 +428,28 @@ def test_enable_fts_error_message_on_views():
         assert e.value.args[0] == "enable_fts() is supported on tables but not on views"
 
 
+def test_enable_fts_twice_without_replace():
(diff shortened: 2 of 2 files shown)


simonw:
I fixed this in https://github.com/simonw/sqlite-utils/commit/1a28416e1095f2b8015195db7f8273b6f7953ab6 instead.

PR #710, opened 2026-02-08
Fix: CSV insert with --detect-types crashes on header-only CSV

### LINKED ISSUE #702: Insert CSV with --detect-types crashes for CSV files that contain only a header row

RamiNoodle733:
## Problem
When using `sqlite-utils insert` with `--detect-types` flag on a CSV file containing only a header row (no data), the command crashes with:
```
```

## Root Cause
The `transform()` method was being called to apply detected types to the table, but when there are no data rows, the table is never created.

## Fix
Added existence checks before calling `transform()` in two locations in `cli.py`.

## Testing
Added `test_insert_csv_header_only()` to verify the fix.

Fixes #702

----
📚 Documentation preview 📚: https://sqlite-utils--710.org.readthedocs.build/en/710/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1176,7 +1176,7 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        if tracker is not None and db.table(table).exists():
             db.table(table).transform(types=tracker.types)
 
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in:
- #710

PR #714, opened 2026-03-06
docs: clarify that insert()/insert_all() commit in batches

### LINKED ISSUE #712: Table.insert() documentation does not make it clear it will commit
[Table.insert()](https://sqlite-utils.datasette.io/en/stable/reference.html#sqlite_utils.db.Table.insert) documentation does not state it will commit any on-going transaction. This [happens inside insert_chunks()](https://github.com/simonw/sqlite-utils/blob/e66e82f2c8b3835a316de6ed1407f1a345358e41/sqlite_utils/db.py#L3285) when using the sqlite connection as a context manager.

This turned out to be an issue in my code where I use manual transaction control, where `insert()` commits the transaction earlier than I expected.

RamiNoodle733:
## Summary
- add a note to the Python API docs for .insert() / .insert_all()
- clarify that writes are committed in batches and may commit earlier than expected in manually-managed transactions

## Testing
- Not run locally (pytest is not installed in this environment).

Fixes #712

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -528,6 +528,12 @@ You can also specify a primary key by passing the ``pk=`` parameter to the ``.in
 
 After inserting a row like this, the ``dogs.last_rowid`` property will return the SQLite ``rowid`` assigned to the most recently inserted record.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


simonw:
Made obsolete by implementation and docs shipped in 4.0.

PR #715, opened 2026-03-07
Avoid loading setuptools plugins under pytest

### LINKED ISSUE #713: Plugins are still loaded when running tests
This code has the intention of only loading plugins when not running tests:

https://github.com/simonw/sqlite-utils/blob/8d74ffc93292c604d5827e2b44fffedca0c28c19/sqlite_utils/plugins.py#L10-L12

https://github.com/simonw/sqlite-utils/blob/8d74ffc93292c604d5827e2b44fffedca0c28c19/tests/conftest.py#L11-L14

However, as far as I can tell, it doesn't function as intended. Plugins are still loaded when running tests. For example, if the https://github.com/simonw/sqlite-migrate plugin is installed on the system, then running pytest fails with errors like this:

```
    def test_register_commands():
        importlib.reload(cli)
>       assert plugins.get_plugins() == []
E       AssertionError: assert [{'hooks': ['...on': '0.1b0'}] == []
E         
E         Left contains one more item: {'hooks': ['register_commands'], 'name': 'sqlite-migrate', 'version': '0.1b0'}
E         Use -v to get more diff

tests/test_plugins.py:9: AssertionError
```

As a result, downstream packaging is more difficult, e.g. https://aur.archlinux.org/packages/sqlite-utils#comment-1031481.

The error appears to be because `getattr(sys, "_called_from_test", False)` is called _while loading_ `sqlite_utils.plugins`, not at runtime. However, `sqlite_utils.plugins` is loaded by `sqlite_utils.db`, which is loaded by `sqlite_utils`, which is loaded by `conftest`, so plugins are already loaded before `pytest_configure` is defined by `conftest`.

RamiNoodle733:
## Summary
Fixes plugin auto-loading during tests when `sys._called_from_test` has not yet been set.

## What changed
- Added `_should_load_setuptools_plugins()` in `sqlite_utils/plugins.py`
- Updated plugin auto-load guard to also check if `pytest` is present in `sys.modules`
- Added regression test for that guard in `tests/test_plugins.py`

## Testing
- `uv run pytest tests/test_plugins.py -q`

Fixes #713

----
📚 Documentation preview 📚: https://sqlite-utils--715.org.readthedocs.build/en/715/

### DIFF
--- a/sqlite_utils/plugins.py
+++ b/sqlite_utils/plugins.py
@@ -7,8 +7,15 @@ from . import hookspecs
 pm: pluggy.PluginManager = pluggy.PluginManager("sqlite_utils")
 pm.add_hookspecs(hookspecs)
 
-if not getattr(sys, "_called_from_test", False):
-    # Only load plugins if not running tests
+
+def _should_load_setuptools_plugins() -> bool:
+    # Avoid loading setuptools plugins while running test suites.
+    # Checking sys.modules for pytest makes this robust even if
+    # sys._called_from_test is set later during test initialization.
+    return not getattr(sys, "_called_from_test", False) and "pytest" not in sys.modules
+
+
+if _should_load_setuptools_plugins():
     pm.load_setuptools_entrypoints("sqlite_utils")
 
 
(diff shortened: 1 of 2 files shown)


simonw:
Handled in:
- #719

PR #716, opened 2026-03-07
fix: Insert CSV with --detect-types crashes for CSV files that contain only a header row

### LINKED ISSUE #702: Insert CSV with --detect-types crashes for CSV files that contain only a header row

RamiNoodle733:
## Summary
Fixes a crash in `sqlite-utils insert` when importing a CSV that only contains a header row with `--detect-types`. The command should succeed without attempting to transform a table that was never created.

## Changes
- Added a guard in the CLI insert flow to only run `transform(types=...)` when the table actually exists
- Added a regression test covering `insert --csv --detect-types` with header-only CSV input

## Testing
- Ran: `.venv/bin/pytest -q tests/test_cli.py -k "detect_types and (insert or upsert)"`
- Result: `9 passed`

Fixes simonw/sqlite-utils#702

----
📚 Documentation preview 📚: https://sqlite-utils--716.org.readthedocs.build/en/716/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1176,7 +1176,7 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        if tracker is not None and db.table(table).exists():
             db.table(table).transform(types=tracker.types)
 
         # Clean up open file-like objects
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in:
- #710

PR #718, opened 2026-03-07
fix: Table.insert() documentation does not make it clear it will commit

### LINKED ISSUE #712: Table.insert() documentation does not make it clear it will commit

RamiNoodle733:
## Summary
Clarifies the `Table.insert()` reference documentation to explicitly state that it commits changes as part of an internal transaction.

## Changes
- Updated the `Table.insert()` docstring in `sqlite_utils/db.py` with a sentence describing commit behavior.

## Testing
- `python3 -m compileall -q sqlite_utils/db.py`
- Could not run `pytest` in this environment (`No module named pytest`).

Closes #712

----
📚 Documentation preview 📚: https://sqlite-utils--718.org.readthedocs.build/en/718/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -3382,6 +3382,8 @@ class Table(Queryable):
         Insert a single record into the table. The table will be created with a schema that matches
         the inserted record if it does not already exist, see :ref:`python_api_creating_tables`.
 
+        This method commits its changes as part of an internal transaction.
+
         - ``record`` - required: a dictionary representing the record to be inserted.
 
         The other parameters are optional, and mostly influence how the new table will be created if


simonw:
Fixed here: https://sqlite-utils.datasette.io/en/latest/python-api.html#transactions-and-saving-your-changes

PR #720, opened 2026-03-08
Fix #699: make square-bracket access return tables only

### LINKED ISSUE #699: Make square braces only return tables, not views

RamiNoodle733:
Fixes #699

## Summary
- Make square-bracket lookup return tables only.
- Views should be accessed explicitly via db.view(name).
- Update CLI code paths that require views accordingly.
- Preserve drop-view ignore behavior and memory command checks.

## Files changed
- sqlite_utils/db.py
- sqlite_utils/cli.py
- tests/test_create.py

## Validation
- pytest tests/test_create.py (drop_view/drop_ignore/bad_table_and_view_exceptions)
- pytest tests/test_cli.py (drop_view)
- pytest tests/test_cli_memory.py
- flake8 on touched files

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1796,7 +1796,10 @@ def drop_view(path, view, ignore, load_extension):
     _register_db_for_cleanup(db)
     _load_extensions(db, load_extension)
     try:
-        db[view].drop(ignore=ignore)
+        db.view(view).drop(ignore=ignore)
+    except sqlite_utils.db.NoView:
+        if not ignore:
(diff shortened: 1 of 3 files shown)


simonw:
I decided to go the other way, square braces still return both, but documentation encourages `db.table()` instead.

PR #721, opened 2026-03-08
Improve typing for db["table"] usage

### LINKED ISSUE #638: Mypy & Pyright show errors for basic sqlite-utils usage
Duplicate of the unresolved closed bug #607.

Here's perfectly valid `sqlite_utils` code that raises objections from Mypy & Pyright/PyLance in VS Code:

```python
#! /usr/bin/env python3

# Create or connect to the database
db = Database("example.db")

# Create a table with an autoincrementing primary key
"""
Mypy/Pylance objects to the create() call with the error:
    "Cannot access attribute "create" for class "View"
Attribute "create" is unknownPylancereportAttributeAccessIssue
"""
db["my_table"].create(
    {"id": int, "name": str, "value": int}, pk="id", if_not_exists=True
)

"""
Mypy/Pylance objects to the upsert() call with the same error.
It also objects to the pk="id" argument, with error:
    "Argument of type "Literal['id']" cannot be assigned to parameter "pk" of type "Default"
    "Literal['id']" is incompatible with "Default"
"""
db["my_table"].upsert({"name": "example", "value": 42}, pk="id")

# Verify the data
print(list(db["my_table"].rows))
```

Is there a workaround or another approach that wouldn't throw these spurious errors?

RamiNoodle733:
## Summary
- annotate `Database.__getitem__` as returning `Table` for static type checkers
- keep existing runtime behavior for views by casting the view return path
- add typing-focused regression tests for the return annotation and view runtime behavior

## Why
Fixes #638, where mypy/pyright report errors for common `db["name"].create(...)` and `upsert(...)` usage.

## Validation
- `uv run flake8 sqlite_utils/db.py tests/test_typing.py`
- `uv run mypy sqlite_utils tests`
- `uv run pytest tests/test_typing.py -q`
- `uv run pytest tests/test_create.py -k view -q`

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -445,7 +445,7 @@ class Database:
         finally:
             self._tracer = prev_tracer
 
-    def __getitem__(self, table_name: str) -> Union["Table", "View"]:
+    def __getitem__(self, table_name: str) -> "Table":
         """
         ``db[table_name]`` returns a :class:`.Table` object for the table with the specified name.
         If the table does not exist yet it will be created the first time data is inserted into it.
@@ -453,7 +453,7 @@ class Database:
         :param table_name: The name of the table
         """
         if table_name in self.view_names():
-            return self.view(table_name)
+            return cast("Table", self.view(table_name))
         return self.table(table_name)
 
     def __repr__(self) -> str:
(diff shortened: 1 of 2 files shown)


simonw:
In v4 you can instead use `db.table(name)` or `db.view(name)` for predictable typing. You can still use `db[name]` for the old behavior but documentation no longer promotes that.

PR #722, opened 2026-03-08
Fix #624: preserve duplicate output column names in query results

### LINKED ISSUE #624: Query with duplicate output column names drop/overwrites duplicate colums
Consider this join query:

    select d1.*, d2.* from delta d1 inner join delta d2 where d1.new==d2.old

This will result in column names from d1, d2 to be duplicated. Unfortunately, running this query via `sqlite-utils`, only one set of output columns is returned, probably being overridden in the map.

RamiNoodle733:
Fixes:
- #624

## Summary
When a query returns duplicate column names (for example joining two tables with the same field names), `Database.query()` and related helpers used `dict(zip(...))`, which silently overwrote earlier values.

This change preserves all values by suffixing duplicate keys as `_2`, `_3`, etc.

## Changes
- Added `_row_to_dict()` helper to safely map row values to unique dictionary keys
- Updated `Database.query()`, `Queryable.rows_where()`, and `Table.search()` to use it
- Added regression test covering duplicate output columns

## Validation
- `pytest -q tests/test_query.py tests/test_rows.py`
- `ruff check sqlite_utils/db.py tests/test_query.py`

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -82,6 +82,23 @@ def quote_identifier(identifier: str) -> str:
     return '"{}"'.format(identifier.replace('"', '""'))
 
 
+def _row_to_dict(keys: Sequence[str], row: Sequence[Any]) -> Dict[str, Any]:
+    """
+    Convert a row plus column names to a dictionary.
+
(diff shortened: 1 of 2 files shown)


simonw:
Fixed here instead:
- #773

PR #730, opened 2026-04-21
fix(cli): avoid transform() when table missing (header-only CSV + detect-types)

### LINKED ISSUE #702: Insert CSV with --detect-types crashes for CSV files that contain only a header row

beejak:
## Summary

Fixes #702.

When importing CSV with default `--detect-types`, a **header-only** file (no data rows) yields an empty insert. The table may not exist yet, but the CLI still called `Table.transform()` from the type tracker path, which asserts that the table exists.

Guard both code paths with `db[table].exists()` before calling `transform()`:

- `insert_upsert_implementation` (e.g. `sqlite-utils insert ... --csv`)
- the multi-file `memory` command path that uses the same tracker + `insert_all` + `transform` sequence

## Test

Added `test_insert_csv_header_only_detect_types_no_crash`: header-only CSV + default detection exits 0 and does not create a table (same net behavior as before for zero rows, without crashing).

## Verification

```bash
pytest tests/test_cli.py::test_insert_csv_header_only_detect_types_no_crash -v
```

----
📚 Documentation preview 📚: https://sqlite-utils--730.org.readthedocs.build/en/730/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1176,7 +1176,7 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        if tracker is not None and db[table].exists():
             db.table(table).transform(types=tracker.types)
 
         # Clean up open file-like objects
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in:
- #710

PR #731, opened 2026-04-22
fix: auto-commit delete_where()

### LINKED ISSUE #159: delete_where does not commit...

cloudyun888:
## Summary
- wrap `Table.delete_where()` in the same connection context manager used by `delete()` and `update()`
- make deletes visible to a new connection immediately instead of leaving them uncommitted
- add a file-backed regression test that reproduces the missing auto-commit behavior

## Testing
- `.venv/bin/python -m pytest tests/test_delete.py -q`
- `.venv/bin/python -m py_compile sqlite_utils/db.py tests/test_delete.py`
- manual repro: delete rows from a file-backed database, reopen it, verify row count stays at 0

Fixes #159

----
📚 Documentation preview 📚: https://sqlite-utils--731.org.readthedocs.build/en/731/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2905,9 +2905,10 @@ class Table(Queryable):
         sql = "delete from {}".format(quote_identifier(self.name))
         if where is not None:
             sql += " where " + where
-        self.db.execute(sql, where_args or [])
-        if analyze:
-            self.analyze()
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in the latest RC.

PR #733, opened 2026-05-04
Guard transform when --detect-types CSV has only a header row (closes #702)

### LINKED ISSUE #702: Insert CSV with --detect-types crashes for CSV files that contain only a header row

guzenkopavel:
Fixes #702.

When `sqlite-utils insert --csv --detect-types` is given a CSV file that contains only the header row (no data rows), the table is never created — but the post-insert call to `db.table(table).transform(types=tracker.types)` still runs and trips an assertion:

This adds the `db.table(table).exists()` guard suggested by the issue author (@nis-djom), plus a regression test (`tests/test_cli.py::test_insert_detect_types_header_only_csv`) that exercises the exact scenario from the issue body.

Local `pytest -q` on the patched branch: 1041 passed, 16 skipped, 6.40s.

----
📚 Documentation preview 📚: https://sqlite-utils--733.org.readthedocs.build/en/733/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1176,7 +1176,7 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        if tracker is not None and db.table(table).exists():
             db.table(table).transform(types=tracker.types)
 
         # Clean up open file-like objects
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in:
- #710

PR #736, opened 2026-05-17
Don't transform empty CSV table that was never created

### LINKED ISSUE #702: Insert CSV with --detect-types crashes for CSV files that contain only a header row

c-tonneslan:
Closes #702.

When a CSV file is header-only, \`insert_all\` is a no-op so the table never gets created, but the \`--detect-types\` path still calls \`transform(types=tracker.types)\` afterwards and \`transform\` asserts \`table.exists()\`. Result is the \`AssertionError: Cannot transform a table that doesn't exist yet\` traceback from the issue.

Gate the \`transform\` call on \`table.exists()\` so the empty case is a clean no-op. Applied the same guard to the analogous loop further down in \`memory\`-style multi-file insertion. Added a regression test that fails on main and passes here.

----
📚 Documentation preview 📚: https://sqlite-utils--736.org.readthedocs.build/en/736/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1175,7 +1175,9 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        if tracker is not None and db.table(table).exists():
+            # A header-only CSV creates no table, so there is nothing to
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in the latest RC.

PR #739, opened 2026-05-17
insert: don't crash on header-only CSV with --detect-types

### LINKED ISSUE #702: Insert CSV with --detect-types crashes for CSV files that contain only a header row

c-tonneslan:
When the CSV has a header but no data rows, the insert path never creates the table — the row loop has nothing to feed into the first insert. The post-insert \`tracker.transform()\` call then asserts the table exists and blows up:

Reported in #702 with the exact fix in the issue body. Guarded the transform on \`db[table].exists()\` so an empty CSV becomes a no-op — matching the behavior for a zero-byte CSV (no header either).

Added a parameterized test (both \`-d\` and \`--detect-types\`) so we don't regress.

Closes #702.

----
📚 Documentation preview 📚: https://sqlite-utils--739.org.readthedocs.build/en/739/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -1175,7 +1175,10 @@ def insert_upsert_implementation(
                 )
             else:
                 raise
-        if tracker is not None:
+        # An empty CSV (header row but no data rows) never creates the table,
+        # so there's nothing to transform. Without this guard, --detect-types
+        # crashes on header-only input (see #702).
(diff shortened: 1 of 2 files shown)


c-tonneslan:
Closing as a duplicate of #736 which I'd already filed earlier. #736 is more complete (covers both the insert path and the memory path).

PR #740, opened 2026-05-18
cli: honor --no-headers for table and tabulate-style formats

### LINKED ISSUE #566: `--no-headers` doesn't work on most formats
Version 3.33

```
sqlite-utils query library.db 'select asin from audible' --fmt plain --no-headers | head -3
asin
0062804006
0062891421
```

c-tonneslan:
Closes #566.

`--no-headers` was already respected for `csv`/`tsv` output, but the two `tabulate.tabulate` calls (used for `--table` and `--fmt`) passed `headers=headers` unconditionally, so the header row still printed. Pass `()` instead when `--no-headers` is set.

----
📚 Documentation preview 📚: https://sqlite-utils--740.org.readthedocs.build/en/740/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -236,7 +236,8 @@ def tables(
             yield row
 
     if table or fmt:
-        print(tabulate.tabulate(_iter(), headers=headers, tablefmt=fmt or "simple"))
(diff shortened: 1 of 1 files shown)


simonw:
This is already fixed.

Issue #625, opened 2024-04-03
Doublebyte content is unicode-escaped in JSON output


oli:
test.json
```json
[
    {
        "id": 123,
        "text": "Japanese 日本語"
    },
    {
        "id": 223,
        "text": [
            "Japanese 日本語"
        ]
    }
]
```

Import
```shell
sqlite-utils insert test.db text test.json --pk=id
```

Query using sqlite-utils
```shell
sqlite-utils test.db "select * from text"
[{"id": 123, "text": "Japanese \u65e5\u672c\u8a9e"},
 {"id": 223, "text": "[\"Japanese \u65e5\u672c\u8a9e\"]"}]

sqlite-utils test.db "select * from text" --csv
id,text
123,Japanese 日本語
223,"[""Japanese 日本語""]"
```

I assume this is something to do with ensure ascii on JSON output, but is there any way I can get non-escaped strings in sqlite-utils JSON output? Thanks!

PR #741, opened 2026-05-18
cli: output JSON without escaping non-ASCII characters

c-tonneslan:
Closes:
- #625

`json.dumps` defaults to `ensure_ascii=True`, so `output_rows` emitted `日本語` instead of `日本語` for Japanese rows. The CSV path already shows the characters verbatim — the JSON path should too. Pass `ensure_ascii=False` so the bytes match what the CSV output (and the database itself) holds.

----
📚 Documentation preview 📚: https://sqlite-utils--741.org.readthedocs.build/en/741/


simonw:
I think this will break on Windows. I'm going to address this in a slightly different way.

simonw:
Fixed in:
- #777

PR #745, opened 2026-05-22
Fix KeyError in insert_all when pk column is missing from a single record

### LINKED ISSUE #732: Creating tables with missing primary keys works inconsistently
db.table.insert_all throws an error when given a primary key that's missing from the input data only in the case where the number of rows equals 1.

```
import sqlite_utils

for n in [0, 1, 2, 3, 10]:
    db.conn.execute("CREATE TABLE t (a TEXT, b INT, PRIMARY KEY (a, b))")
    
    rows = [{"a": f"x{i}", "b": i} for i in range(n)]
    
    try:
        db.table("t").insert_all(rows, pk="random_text_thats_not_in_the_table", alter=True)
        print(f"n={n}: no error")
    except KeyError as e:
        print(f"n={n}: KeyError {e}")
```

outputs:

```
n=0: no error
n=1: KeyError 'id'
n=2: no error
n=3: no error
n=10: no error
```

Seems like there should either always or never be an error in this case

ATOM00blue:
Fixes #732

## The bug

`insert_all()` with a `pk=` column that isn't present in the records raised a `KeyError` — but only when inserting exactly one row. Every other row count succeeded:

```python
import sqlite_utils

for n in [0, 1, 2, 3, 10]:
    db.conn.execute("CREATE TABLE t (a TEXT, b INT, PRIMARY KEY (a, b))")
    rows = [{"a": f"x{i}", "b": i} for i in range(n)]
    try:
        db.table("t").insert_all(rows, pk="not_a_column", alter=True)
        print(f"n={n}: ok")
    except KeyError as e:
        print(f"n={n}: KeyError {e}")
```

```
n=0: ok
n=1: KeyError 'not_a_column'
n=2: ok
n=3: ok
n=10: ok
```

## Root cause

The single-record branch that populates `last_pk` reads the pk value straight out of the just-inserted row (`row[pk]`). When the named pk column doesn't actually exist on the table, that lookup throws. The multi-record path doesn't go through this branch, hence the inconsistency.

## Fix

Check that the named pk column(s) are present on the row before reading them, and fall back to the rowid otherwise — the same fallback already used when no pk is given. Single-row inserts now behave like every other row count.

## Testing

Added `test_insert_all_pk_not_in_records`, parametrized over 0/1/2/3/10 rows, which fails on `main` (KeyError for the single-row case) and passes with this change. Full test suite, black, flake8 and mypy all pass locally.

----
📚 Documentation preview 📚: https://sqlite-utils--745.org.readthedocs.build/en/745/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -3608,12 +3608,20 @@ class Table(Queryable):
                 if (hash_id or pk) and self.last_rowid:
                     # Set self.last_pk to the pk(s) for that rowid
                     row = list(self.rows_where("rowid = ?", [self.last_rowid]))[0]
-                    if hash_id:
-                        self.last_pk = row[hash_id]
-                    elif isinstance(pk, str):
-                        self.last_pk = row[pk]
+                    pk_cols = (
+                        [hash_id]
+                        if hash_id
+                        else ([pk] if isinstance(pk, str) else list(pk))
+                    )
+                    if all(col in row for col in pk_cols):
+                        if hash_id or isinstance(pk, str):
+                            self.last_pk = row[pk_cols[0]]
+                        else:
+                            self.last_pk = tuple(row[col] for col in pk_cols)
                     else:
-                        self.last_pk = tuple(row[p] for p in pk)
+                        # Named pk column(s) are not present in the table - fall
+                        # back to the rowid, matching the multi-row behaviour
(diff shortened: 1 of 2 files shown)


simonw:
I went with a different solution: https://github.com/simonw/sqlite-utils/commit/6225eba5c8daf97a5e459e9fcce2e46bcd022f50

PR #749, opened 2026-06-02
Fix .transform() failing when the table is part of a view

### LINKED ISSUE #586: .transform() fails to drop column if table is part of a view
> I wonder if this should become an option in `sqlite-utils`? Maybe a `recreate_views=True` argument for `table.tranform(...)`? Should it be opt-in or opt-out?

_Originally posted by @simonw in https://github.com/simonw/datasette-edit-schema/issues/35#issuecomment-1683370548_

viraatdas:
Fixes #586.

`Table.transform()` raised `error in view <name>: no such table: ...` when the table was referenced by a SQL view, because the transform renames the table while a view still references it.

This saves the definitions of all views (and any `INSTEAD OF` triggers on them), drops the views before the transform runs, then recreates them afterwards. If a view cannot be recreated because it referenced a column that the transform dropped or renamed, it raises a clear `TransformError` naming the view and telling the user to recreate it manually, rather than leaving the database in a broken state. It deliberately does not parse the view SQL; it relies on SQLite to report a recreate failure.

Added regression tests in `tests/test_transform.py` covering: a view recreated after a transform, an `INSTEAD OF` trigger on a view recreated, and a view referencing an unregistered custom function (which recreates fine since `CREATE VIEW` is lazy). The transform test file passes (62 tests).

Prepared with AI assistance.

----
📚 Documentation preview 📚: https://sqlite-utils--749.org.readthedocs.build/en/749/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -1905,6 +1905,17 @@ class Table(Queryable):
             column_order=column_order,
             keep_table=keep_table,
         )
+        views = self.db.execute(
+            "select name, sql from sqlite_master where type = 'view' "
+            "and sql is not null order by rowid"
+        ).fetchall()
+        view_triggers = []
+        if views:
+            view_triggers = self.db.execute(
+                "select name, sql from sqlite_master where type = 'trigger' "
+                "and sql is not null and tbl_name in "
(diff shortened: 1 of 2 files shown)


simonw:
Superseded by:
- #832

Issue #186, opened 2020-10-16
.extract() shouldn't extract null values


simonw:
This almost works, but it creates a rogue `type` record with a value of None.
```
In [1]: import sqlite_utils
In [2]: db = sqlite_utils.Database(memory=True)
In [5]: db["creatures"].insert_all([
    {"id": 1, "name": "Simon", "type": None},
    {"id": 2, "name": "Natalie", "type": None},
    {"id": 3, "name": "Cleo", "type": "dog"}], pk="id")
Out[5]: <Table creatures (id, name, type)>
In [7]: db["creatures"].extract("type")
Out[7]: <Table creatures (id, name, type_id)>
In [8]: list(db["creatures"].rows)
Out[8]: 
[{'id': 1, 'name': 'Simon', 'type_id': None},
 {'id': 2, 'name': 'Natalie', 'type_id': None},
 {'id': 3, 'name': 'Cleo', 'type_id': 2}]
In [9]: db["type"]
Out[9]: <Table type (id, type)>
In [10]: list(db["type"].rows)
Out[10]: [{'id': 1, 'type': None}, {'id': 2, 'type': 'dog'}]
```

simonw:
How should this work with extractions covering multiple columns?

If there's a single column then it makes sense that a `null` value would not be extracted into the lookup table, but would instead become stay as `null`.

For a multiple column extraction, provided at least one of those columns is not null It should map to a record in the lookup table. Only if ALL of the extracted columns are null should the lookup value stay null.

simonw:
The alternative solution here would be that a single `null` value DOES get extracted. To implement this I would need to add some logic that uses `is null` instead of `=`.

simonw:
Either way I think I'm going to need to add some SQL which uses `where a = b or (a is null and b is null)`.

simonw:
Actually I think this should be an option to `.extract()` which controls if nulls are extracted or left alone.

simonw:
Failing test:
```python
def test_extract_null_values(fresh_db):
    fresh_db["species"].insert({"id": 1, "species": "Wolf"}, pk="id")
    fresh_db["individuals"].insert_all(
        [
            {"id": 10, "name": "Terriana", "species": "Fox"},
            {"id": 11, "name": "Spenidorm", "species": None},
            {"id": 12, "name": "Grantheim", "species": "Wolf"},
            {"id": 13, "name": "Turnutopia", "species": None},
[...]

PR #750, opened 2026-06-07
Don't extract NULL values into a lookup row (#186)

JOhnsonKC201:
Refs:
- #186

## Problem

`table.extract()` populated its lookup table with:

```sql
INSERT OR IGNORE INTO lookup (cols) SELECT DISTINCT cols FROM source
```

That `SELECT DISTINCT` includes the all-NULL combination, so a spurious lookup row is created for NULL and every NULL source row is pointed at it — instead of keeping a NULL foreign key.

```python
db["creatures"].insert_all([
    {"id": 1, "name": "Simon", "type": None},
    {"id": 2, "name": "Natalie", "type": None},
    {"id": 3, "name": "Cleo", "type": "dog"}], pk="id")
db["creatures"].extract("type")
```

**Before**
```
type:      [{"id": 1, "type": None}, {"id": 2, "type": "dog"}]
creatures: Simon -> type_id=1, Natalie -> type_id=1, Cleo -> type_id=2
```

**After**
```
type:      [{"id": 1, "type": "dog"}]
creatures: Simon -> type_id=None, Natalie -> type_id=None, Cleo -> type_id=1
```

## Fix

A row whose extracted columns are entirely NULL represents "no value", so it should keep a NULL foreign key and not get a lookup row. The lookup `INSERT` now has a `WHERE NOT (<col> IS NULL AND ...)` guard, and the existing `IS`-based foreign-key `UPDATE` then leaves those rows NULL automatically (its subquery finds no matching lookup row).

For multi-column extracts, **only the fully-NULL combination is skipped** — a partial-NULL combination (some extracted columns set, others NULL) is a genuine distinct value and is still extracted and shared between matching rows.

## Tests

- Updated `test_extract_works_with_null_values` to assert the corrected behaviour (NULL stays NULL, no NULL lookup row).
- Added `test_extract_does_not_create_lookup_row_for_all_null` (single column).
- Added `test_extract_multi_column_keeps_partial_null_but_not_all_null` (multi-column partial vs all-NULL).
- Full suite: **1041 passed, 17 skipped** locally.

----
📚 Documentation preview 📚: https://sqlite-utils--750.org.readthedocs.build/en/750/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2267,14 +2267,22 @@ class Table(Queryable):
                 )
             lookup_columns = [(rename.get(col) or col) for col in columns]
             lookup_table.create_index(lookup_columns, unique=True, if_not_exists=True)
+            # Don't create a lookup row for the all-NULL combination: a row whose
+            # extracted columns are entirely NULL represents "no value", so it
+            # should keep a NULL foreign key rather than point at a NULL lookup row
+            # (#186). Rows with a partial NULL (some extracted columns set, others
+            # NULL) are a genuine distinct value and are still extracted.
             self.db.execute(
-                "INSERT OR IGNORE INTO {} ({lookup_columns}) SELECT DISTINCT {table_cols} FROM {}".format(
+                "INSERT OR IGNORE INTO {} ({lookup_columns}) SELECT DISTINCT {table_cols} FROM {} WHERE NOT ({all_null})".format(
                     quote_identifier(table),
                     quote_identifier(self.name),
                     lookup_columns=", ".join(
                         quote_identifier(c) for c in lookup_columns
                     ),
                     table_cols=", ".join(quote_identifier(c) for c in columns),
+                    all_null=" AND ".join(
+                        "{} IS NULL".format(quote_identifier(c)) for c in columns
+                    ),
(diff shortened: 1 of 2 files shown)


JOhnsonKC201:
Good catch on the reuse case — confirmed it reproduces: with a lookup table that already contains an all-NULL row, the `IS`-based foreign-key UPDATE still linked all-NULL source rows to it.

Pushed 89fe8bd: the UPDATE now carries a trailing `WHERE NOT (<col> IS NULL AND ...)`, so all-NULL source rows keep a NULL foreign key regardless of the lookup table's existing contents. Added `test_extract_all_null_stays_null_with_preexisting_null_lookup_row` to cover it. Full suite still green (1042 passed).

JOhnsonKC201:
Friendly ping on this one. The full test suite is still waiting on a maintainer to approve the CI run — it only has the docs checks so far. The change keeps all-NULL source rows from being linked into a lookup row (including the pre-existing-NULL-row case covered by the added test). Happy to address anything once CI runs.

simonw:
Thanks for this, I ended up going with a slightly different fix: 2616dec7957a366da3260cf87eb79c78bae39d14

PR #751, opened 2026-06-08
Honor --no-headers for --fmt and --table output (#566)

JOhnsonKC201:
Refs #566.

## Problem

`--no-headers` was only honored for `--csv`/`--tsv` output. For the tabulate formats (`--fmt ...` and `-t`/`--table`) the flag was **silently ignored**, so the column-name header row was always printed:

```
$ sqlite-utils query library.db 'select asin from audible' --fmt plain --no-headers
asin        # <- should not be here
0062804006
0062891421
```

The two `tabulate.tabulate(...)` call sites (the `query`/`memory` output and the `rows`/`tables` output) always passed `headers=headers` regardless of the flag.

## Fix

Both call sites now pass `headers=() if no_headers else headers`. `tabulate` renders headerless output cleanly for every format (`plain`, `simple`, `github`, `grid`, `rst`, …), so `--no-headers` now works consistently across CSV, TSV and all `--fmt`/`--table` formats. Default output (headers shown) is unchanged, and the CSV/TSV path is untouched.

```
$ sqlite-utils query library.db 'select asin from audible' --fmt plain --no-headers
0062804006
0062891421
```

I went with making the flag do what its name says (rather than raising an error on `--fmt` + `--no-headers`) because `tabulate` supports headerless rendering for all formats and this matches the documented purpose of the flag — and the use case in the issue (piping a single column into `xargs`/unix tools) is a legitimate one.

The `--no-headers` help text previously said "Omit CSV headers"; it now reads "Omit headers from CSV/TSV and table/--fmt output", and `docs/cli-reference.rst` is regenerated via cog to match.

## Tests

Added `test_output_table_no_headers`, parametrized over `--fmt simple`, `-t` and `--fmt github`, asserting the header row is omitted (and the data preserved) for both the `query` and `rows` commands.

- Full suite: **1042 passed, 17 skipped** locally.
- `black . --check`, `cog --check README.md docs/*.rst` and `codespell` all pass.

----
📚 Documentation preview 📚: https://sqlite-utils--751.org.readthedocs.build/en/751/

### DIFF
--- a/docs/cli-reference.rst
+++ b/docs/cli-reference.rst
@@ -115,7 +115,7 @@ See :ref:`cli_query`.
       --arrays                    Output rows as arrays instead of objects
       --csv                       Output CSV
       --tsv                       Output TSV
-      --no-headers                Omit CSV headers
+      --no-headers                Omit headers from CSV/TSV and table/--fmt output
       -t, --table                 Output as a formatted table
       --fmt TEXT                  Table format - one of asciidoc, colon_grid,
                                   double_grid, double_outline, fancy_grid,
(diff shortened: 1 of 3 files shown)


JOhnsonKC201:
Friendly ping — same as #750, the full test suite is waiting on a maintainer CI approval (only the docs checks have run). This makes `--fmt` and `--table` output honor `--no-headers`. Happy to make changes once CI runs.

PR #753, opened 2026-06-10
Raise ValueError instead of assert for conflicting insert_all() options

IgnazioDS:
`insert_all()` validates two mutually-exclusive argument pairs — `pk=`/`hash_id=` and `ignore=`/`replace=` — with bare `assert` statements:

```python
assert not (hash_id and pk), "Use either pk= or hash_id="
...
assert not (ignore and replace), "Use either ignore=True or replace=True, not both"
```

Two problems:

1. `assert` statements are stripped when Python runs under `-O`/`-OO`, so on optimized deployments these checks vanish entirely and a contradictory call (e.g. `insert_all(rows, pk="id", hash_id="hash")`) silently proceeds instead of being rejected.
2. When they do fire they raise `AssertionError`, which is inconsistent with the neighbouring validation a couple of lines above that raises `PrimaryKeyRequired` / `ValueError` for the same kind of argument problem.

This replaces both asserts with explicit `ValueError` raises (same messages) and adds regression tests for the two conflicts. No behaviour change for valid calls.

Full test suite passes (1042 passed, 16 skipped); `black`/`flake8` clean.

----
📚 Documentation preview 📚: https://sqlite-utils--753.org.readthedocs.build/en/753/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -3464,15 +3464,15 @@ class Table(Queryable):
         if upsert and (not pk and not hash_id):
             raise PrimaryKeyRequired("upsert() requires a pk")
 
-        assert not (hash_id and pk), "Use either pk= or hash_id="
+        if hash_id and pk:
+            raise ValueError("Use either pk= or hash_id=")
         if hash_id_columns and (hash_id is None):
             hash_id = "id"
         if hash_id:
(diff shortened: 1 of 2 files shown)


simonw:
Fixed in the latest RC.

Issue #554, opened 2023-05-22
`IndexError` when doing `.insert(..., pk='id')` after `insert_all`


xavdid:
I believe this is related to https://github.com/simonw/sqlite-utils/issues/98.

When `pk` is specified by table A's `insert` call, it throws an index error if a different table has written a row with a higher rowid than exists in the first table. Here's a basic example:

```py
from sqlite_utils import Database


def test_pk_for_insert(fresh_db):
    user = {"id": "abc", "name": "david"}

    fresh_db["users"].insert(user, pk="id")

    fresh_db["comments"].insert_all(
        [
            {"id": "def", "text": "ok"},
            {"id": "ghi", "text": "great"},
        ],
    )

    fresh_db["users"].insert(
        user,
        ignore=True,
        # BUG: when specifying pk on the second insert call 
        # db.py goes into a block it doesn't expect and we get the error
        pk="id",
    )


if __name__ == "__main__":
    db = Database("bug.db")
    if db["users"].exists():
        raise ValueError(
            "bug only shows on a new database - remove bug.db before running the script"
        )
    test_pk_for_insert(db)
```

The error is:

```py
  File "/Users/david/projects/reddit-to-sqlite/.venv/lib/python3.11/site-packages/sqlite_utils/db.py", line 2960, in insert_chunk
    row = list(self.rows_where("rowid = ?", [self.last_rowid]))[0]
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

The issue is in this block: 

https://github.com/simonw/sqlite-utils/blob/2747257a3334d55e890b40ec58fada57ae8cfbfd/sqlite_utils/db.py#L2954-L2958

relevant locals are:

- `pk`: `'id'`
- `result.lastrowid`: `2`

What's most interesting is the comment `# self.last_rowid will be 0 if a "INSERT OR IGNORE" happened`, which doesn't seem to be the case here.

xavdid:
Oh and for context - this goes away if I use `.upsert` instead of `insert(..., ignore=True)`, but I don't want to update the value if it's written, just do an insert if it's new. The code is basically:

```py
def save_items(table, items):
    db["users"].insert(build_user(items[0]), pk="id",ignore=True)
    db[table].insert_all(items)

if comments := fetch_comments():
    save_items('comments', comments)

if posts := fetch_posts():
    save_items('posts', posts)
```

[...]

PR #756, opened 2026-06-21
Handle ignored inserts with explicit primary keys

KairosOps:
﻿## Summary

This fixes the `IndexError` from #554 when a single-row `insert(..., pk=..., ignore=True)` is ignored after another table has advanced SQLite's connection-level `lastrowid`.

In that case `cursor.lastrowid` can point at a rowid from a different table, so looking up that rowid in the current table can return no rows. The insert already had an explicit primary key, so when SQLite reports `rowcount == 0` for the ignored insert this now keeps `last_pk` based on the input record instead of trying to resolve it through the stale rowid.

## Checks

- `python -m pytest tests/test_create.py::test_insert_ignore tests/test_create.py::test_insert_ignore_with_pk_after_other_table_insert -q`
- `python -m pytest tests/test_create.py -q`
- `python -m py_compile sqlite_utils/db.py tests/test_create.py`
- `git diff --check`
- `python -m black --fast --check sqlite_utils/db.py tests/test_create.py`

Refs #554

----
📚 Documentation preview 📚: https://sqlite-utils--756.org.readthedocs.build/en/756/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -3604,18 +3604,42 @@ class Table(Queryable):
         if num_records_processed == 1:
             # For an insert we need to use result.lastrowid
             if not upsert and result is not None:
-                self.last_rowid = result.lastrowid
-                if (hash_id or pk) and self.last_rowid:
-                    # Set self.last_pk to the pk(s) for that rowid
-                    row = list(self.rows_where("rowid = ?", [self.last_rowid]))[0]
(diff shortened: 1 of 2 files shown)


simonw:
I fixed this in https://github.com/simonw/sqlite-utils/commit/221774f25a53c76874b28af1c2dd170c76bbd9f1

PR #764, opened 2026-06-30
Fix transform() corrupting TRUE/FALSE/NULL column defaults into strings

gaoflow:
### The bug

`table.transform()` corrupts `TRUE` / `FALSE` / `NULL` keyword-literal column defaults into string literals, silently changing the value future rows receive.

```python
import sqlite_utils
db.execute("CREATE TABLE t (id INTEGER PRIMARY KEY, is_active INTEGER DEFAULT TRUE)")
db["t"].insert({"id": 1})
db["t"].transform(rename={"id": "pk"})          # any structure-only rebuild
print(db["t"].schema)                            # ... "is_active" INTEGER DEFAULT 'TRUE'
db["t"].insert({})
print(db.execute("SELECT is_active FROM t").fetchall())   # [(1,), ('TRUE',)]  <-- corrupted
```

Before the transform a default insert stored the integer `1`; afterwards it stores the text `'TRUE'`. Same for `FALSE` (→ `'FALSE'` instead of `0`) and `NULL` (→ `'NULL'` instead of null). On an `INTEGER` column the quoted `'TRUE'` isn't numeric, so it's stored as TEXT — a genuine type and value change.

### Cause

`transform()` rebuilds the table by re-emitting each column's `PRAGMA table_info` default through `quote_default_value()`. That method returns the value verbatim when it is already quoted, is a `CURRENT_TIME/DATE/TIMESTAMP` literal, or ends with `)`; otherwise it calls `self.quote()`. The keyword literals `TRUE`, `FALSE` and `NULL` match none of those guards, so they get string-quoted.

### The fix

Return `TRUE` / `FALSE` / `NULL` unquoted, mirroring the existing `CURRENT_*` handling. A populated/numeric/expression default is unaffected.

### Tests

Added `TRUE` / `FALSE` / `true` / `NULL` cases to the `quote_default_value` `EXAMPLES` table, and a functional `test_transform_preserves_keyword_literal_defaults` that creates `DEFAULT TRUE/FALSE/NULL`, rebuilds via `transform()`, and asserts the schema stays unquoted **and** a fresh default insert still yields `1 / 0 / NULL`. All new cases fail on `main` and pass with the fix; the full `test_default_value.py` + `test_transform.py` suites stay green (78 passed); `black`, `flake8` and `mypy` clean.

----
📚 Documentation preview 📚: https://sqlite-utils--764.org.readthedocs.build/en/764/

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -42,6 +42,10 @@ Column names passed to Python API methods are now matched against the table sche
 - Foreign key columns are validated and recorded using the casing of the actual schema columns, in ``foreign_keys=`` when creating tables, ``db.add_foreign_keys()``, ``table.add_foreign_key()`` and ``table.add_column(fk_col=...)``. Duplicate foreign key detection is also case-insensitive.
 - ``table.create()`` with ``pk=``, ``not_null=``, ``defaults=`` or ``column_order=`` referencing columns using different casing no longer creates an unwanted extra primary key column or raises a ``ValueError``.
 
+Everything else:
+
+- Fixed a bug where ``table.transform()`` could convert ``DEFAULT TRUE``, ``DEFAULT FALSE`` and ``DEFAULT NULL`` column defaults into quoted string defaults when rebuilding a table. Thanks, `Vincent Gao <https://github.com/gaoflow>`__. (`#764 <https://github.com/simonw/sqlite-utils/pull/764>`__)
+
 .. _v4_0rc2:
 
 4.0rc2 (2026-07-04)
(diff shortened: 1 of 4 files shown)


simonw:
Looks good, introduces one new test failure though:
```
    @pytest.mark.parametrize(
        "value",
        (
            1,
            1.3,
            "foo",
            True,
            b"binary",
        ),
    )
    def test_table_default_values(fresh_db, value):
        fresh_db["default_values"].insert(
            {"nodefault": 1, "value": value}, defaults={"value": value}
        )
        default_values = fresh_db["default_values"].default_values
>       assert default_values == {"value": value}
E       AssertionError: assert {'value': 'True'} == {'value': True}
E         
E         Differing items:
E         {'value': 'True'} != {'value': True}
E         Use -v to get more diff
```

simonw:
Thanks for this!

PR #766, opened 2026-07-01
Preserve CHECK constraints when transforming a table

### LINKED ISSUE #762: .transform() drops check constraints and comments
Calling `.transform()` on a table currently recreates that table without check constraints and without any comments, because those live only in the SQL create table statement and are not exposed via any PRAGMA.

Separately, if the table has a column-level `UNIQUE` constraint, SQLite backs it with an internal auto-index that has no `CREATE INDEX` statement. `.transform()`'s index-rebuild step can't reproduce that, so it raises `TransformError` and aborts instead of preserving the constraint.

apoorva-01:
`.transform()` quietly drops a table's CHECK constraints. They only live in the stored `CREATE TABLE` SQL, and transform rebuilds from PRAGMA metadata, which doesn't expose them.

This pulls the CHECK expressions out of the schema and re-adds them. A rename rewrites the columns a constraint references, and dropping a referenced column drops its constraint, so transform never emits a table that won't build. Column-level checks come back as table-level (identical in SQLite). It uses a small quote/paren-aware tokenizer, so no new dependency; the one edge it can't see is a column whose name also appears as a function call (`length` alongside `length(x)`), left as a documented limitation.

I scoped this to CHECK constraints. Comments and column-level UNIQUE are follow-ups (comment placement gets ambiguous on rename/reorder; UNIQUE is the separate auto-index case), happy to take either. Tests cover round-trip, rename, drop and parser edges and fail without the fix. Didn't touch the changelog since you compile it at release.

Closes #762

----
📚 Documentation preview 📚: https://sqlite-utils--766.org.readthedocs.build/en/766/

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -1505,6 +1505,8 @@ To keep the original table around instead of dropping it, pass the ``keep_table=
 
     table.transform(types={"age": int}, keep_table="original_table")
 
+``CHECK`` constraints are preserved across a transform. If a column referenced by a constraint is renamed the constraint is updated to match, and if such a column is dropped the constraint is dropped along with it.
+
 This method raises a ``sqlite_utils.db.TransformError`` exception if the table cannot be transformed, usually because there are existing constraints or indexes that are incompatible with modifications to the columns.
 
 .. _python_api_transform_alter_column_types:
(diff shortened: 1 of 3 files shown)


simonw:
This is now fixed on main in a slightly different way.

PR #773, opened 2026-07-05
Preserve duplicate column names in query results, closes #624

### LINKED ISSUE #624: Query with duplicate output column names drop/overwrites duplicate colums

simonw:
Queries returning duplicate column names - e.g. joins between tables sharing column names - silently lost values because rows were built with dict(zip(keys, row)), where the last duplicate wins.

Later occurrences are now renamed with a numeric suffix: id, id becomes id, id_2 - skipping any suffix that would collide with a real column in the same query.

The new utils.dedupe_keys() helper transforms the key list once per query, so the per-row dict construction is unchanged and there is no measurable performance impact.

Applied in Database.query() (including the PRAGMA and RETURNING paths), Table.rows_where(), Table.search() and the CLI's JSON output. CSV, TSV and table output keep the original duplicate headers.

Replaces:
- #722

Closes:
- #624

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -45,6 +45,8 @@ The default format returned for queries is JSON:
     [{"id": 1, "age": 4, "name": "Cleo"},
      {"id": 2, "age": 2, "name": "Pancakes"}]
 
+If the query returns more than one column with the same name, later occurrences are renamed with a numeric suffix - ``select 1 as id, 2 as id`` returns ``[{"id": 1, "id_2": 2}]``. This only applies to JSON output: :ref:`CSV and TSV <cli_query_csv>` and :ref:`table <cli_query_table>` output keep the duplicate column headers unchanged.
+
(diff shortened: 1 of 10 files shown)


simonw:
Without this fix:
```bash
uvx sqlite-utils memory 'select 1 as n, 2 as n'
```
```json
[{"n": 2}]
```
With this fix:
```bash
uv run sqlite-utils memory 'select 1 as n, 2 as n'
```
```json
[{"n": 1, "n_2": 2}]
```

simonw:
Failures in CI relate to the GitHub Actions cache service, I'm merging anyway.

Issue #781, opened 2026-07-07
Regression: rowid primary key rejected by upsert(pk=)


simonw:
Refs:
- https://github.com/simonw/sqlite-utils/issues/769#issuecomment-4900585395

STR:
```python
import sqlite_utils
db = sqlite_utils.Database(memory=True)
db["t"].insert({"title": "Hello"})     # rowid table, columns = ['title']
db["t"].pks                             # -> ['rowid']
db["t"].upsert({"rowid": 1, "title": "x"}, pk="rowid")
# 4.0rc4: InvalidColumns: Invalid primary key column ['rowid'] for table t with columns ['title']
# 3.39:   works

Issue #769, opened 2026-07-05
Release sqlite-utils 4.0 stable


simonw:
After the two RCs I think this is pretty well baked. I've been running test suites from `datasette` and `llm` against it which are giving me confidence it's ready to go.

The only slightly tricky bit here is co-ordinating the release of `sqlite-migrate`. Current plan:
- Release a `sqlite-migrate` version that pins to `sqlite-utils<4`
- Release v4
- Release a new `sqlite-migrate` version that pins to `sqlite-utils>=4.0` and eliminates all the code, just doing `from sqlite_utils import Migrations`. https://github.com/simonw/sqlite-migrate/issues/18

simonw:
Refs:
- https://github.com/simonw/sqlite-utils/issues/758

simonw:
Some issues to resolve before the 4 stable release, as fixing them later would require a 5 release:
- [x] https://github.com/simonw/sqlite-utils/issues/705
  - [x] https://github.com/simonw/sqlite-utils/pull/776
- [x] https://github.com/simonw/sqlite-utils/issues/760
- [x] Maybe compound foreign keys?
  - https://github.com/simonw/sqlite-utils/issues/117
  - https://github.com/simonw/sqlite-utils/issues/594

simonw:
On compound foreign keys, I don't think I'm going to add support for creating them BUT issue #594 is about how they are currently incorrectly represented in `table.foreign_keys` output - as two separate keys, with the position information lost - which feels bad.

simonw:
Landed quite comprehensive compound foreign key support in:
- #770 

Including creating them!

simonw:
There's enough in here that I think we should do another RC, so I can easily test it in `datasette` and `llm`.

simonw:
More things to land before 4.0 final:
- [x] https://github.com/simonw/sqlite-utils/pull/751
- [x] https://github.com/simonw/sqlite-utils/pull/741
- [x] https://github.com/simonw/sqlite-utils/pull/750
- [x] https://github.com/simonw/sqlite-utils/pull/745
- [x] https://github.com/simonw/sqlite-utils/pull/756

simonw:
LLM test suite passed with rc3 - https://github.com/simonw/llm/actions/runs/28770652532/job/85303561750

[...]

PR #782, opened 2026-07-07
Fix rowid pk and last_rowid regressions in insert/upsert

simonw:
Refs:
- #783 
- #781
- #769 

Two behaviour regressions in the 4.0 insert/upsert rewrite broke callers
(notably Datasette's write API) that operate on tables without an explicit
primary key. Both are fixed here with regression tests.

1. rowid (and its aliases _rowid_/oid) were rejected as a primary key.
   Table.pks already reports ["rowid"] for a rowid table, but the new pk
   validation raised InvalidColumns because rowid is not listed among the
   table's columns, and the insert success path then raised KeyError when
   looking up the pk value. rowid aliases are now accepted for rowid tables
   and resolve directly to the rowid.

2. An ignored insert (INSERT OR IGNORE that matched an existing row) no
   longer populated last_rowid, and only set last_pk when an explicit pk=
   was passed. It now locates the existing conflicting row by its primary
   key values and reports that row's rowid and pk, rather than relying on
   the connection's last inserted rowid.

Add a shared ROWID_ALIASES constant for the rowid alias names.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01E7af8SxFZqiCerJB6MqKnY

----
📚 Documentation preview 📚: https://sqlite-utils--782.org.readthedocs.build/en/782/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -4532,9 +4532,7 @@ class Table(Queryable):
                     else:
                         first_record_dict = cast(Dict[str, Any], first_record)
                     if hash_id:
-                        self.last_pk = hash_record(
-                            first_record_dict, hash_id_columns
-                        )
+                        self.last_pk = hash_record(first_record_dict, hash_id_columns)
                     elif isinstance(pk, str):
                         self.last_pk = first_record_dict[
                             resolve_casing(pk, first_record_dict)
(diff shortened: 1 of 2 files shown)


simonw:
That last commit is because I prompted:

> `The change in db.py is quite complex check if the tests fully cover that using coverage`

PR #785, opened 2026-07-08
Fix transform to convert empty strings to NULL when changing to integer or float type

### LINKED ISSUE #488: `sqlite-utils transform` should set empty strings to null when converting text columns to integer/float
```
/tmp % echo "id,age,weight\n1,3,2.5\n2,," | sqlite-utils insert test.db test - --csv
/tmp % sqlite-utils schema test.db                                                  
CREATE TABLE [test] (
   [id] TEXT,
   [age] TEXT,
   [weight] TEXT
);
/tmp % sqlite-utils transform test.db test --type age integer --type weight float   
/tmp % sqlite-utils schema test.db                                               
CREATE TABLE "test" (
   [id] TEXT,
   [age] INTEGER,
   [weight] FLOAT
);
/tmp % sqlite-utils rows test.db test
[{"id": "1", "age": 3, "weight": 2.5},
 {"id": "2", "age": "", "weight": ""}]
```
It would be neat if this resulted in the following instead:
```
 {"id": "2", "age": null, "weight": null}
```
Related Discord discussion: https://discord.com/channels/823971286308356157/823971286941302908/1019635490833567794

ikatyal2110:
When you run `.transform(types={"age": int})` on a table that has empty strings in a TEXT column, the empty strings currently survive as `""` in the new INTEGER/FLOAT column instead of becoming NULL.

The fix wraps the SELECT expression with `NULLIF(col, '')` for any column explicitly being converted to a numeric type, so empty strings become NULL during the data copy step.

Text columns and columns not listed in `types` are not affected.

Fixes #488

----
📚 Documentation preview 📚: https://sqlite-utils--785.org.readthedocs.build/en/785/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2820,10 +2820,24 @@ class Table(Queryable):
         if "rowid" not in new_cols:
             new_cols.insert(0, "rowid")
             old_cols.insert(0, "rowid")
+        # Columns explicitly converted to a numeric type need NULLIF(col, '') so
+        # that empty strings stored in a previously TEXT column become NULL rather
+        # than being coerced to 0 or raising a type error.
+        _numeric_kws = ("INT", "REAL", "FLOA", "DOUB", "NUMERIC", "DECIMAL")
+
+        def _col_expr(from_, to_):
+            if from_ in types:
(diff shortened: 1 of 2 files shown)


simonw:
Dupe of:
- #805

PR #786, opened 2026-07-09
Fix .transform() raising TransformError on column-level UNIQUE constraints

### LINKED ISSUE #762: .transform() drops check constraints and comments

ikatyal2110:
When a table has a column-level `UNIQUE` constraint (e.g. `name TEXT UNIQUE`), SQLite backs it with an auto-index whose `sql` column in `sqlite_master` is `NULL`. The index-rebuild step in `transform_sql()` treated any `NULL`-sql index as unrecoverable, raising `TransformError` and aborting the transform entirely. This adds a `unique` parameter to `create_table_sql` so the constraint is reproduced as an inline `UNIQUE` column attribute in the new `CREATE TABLE` statement; the index loop now skips auto-UNIQUE indices (`origin="u"`) instead of raising. Column renames are applied to the constraint, and dropping a UNIQUE column drops its constraint silently. Fixes #762

---
_Generated by [Claude Code](https://claude.ai/code/session_01LPQoysxk76C5tD2CwHWgbr)_

----
📚 Documentation preview 📚: https://sqlite-utils--786.org.readthedocs.build/en/786/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -1400,6 +1400,7 @@ class Database:
         extracts: Optional[Union[Dict[str, str], List[str]]] = None,
         if_not_exists: bool = False,
         strict: bool = False,
+        unique: Optional[Iterable[str]] = None,
     ) -> str:
         """
         Returns the SQL ``CREATE TABLE`` statement for creating the specified table.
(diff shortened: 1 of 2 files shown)


simonw:
Superseded and fixed in 4.2.

Issue #787, opened 2026-07-11
.transform(strict=True/False) and sqlite-utils transform --strict/--no-strict


simonw:
Inspired by https://evanhahn.com/prefer-strict-tables-in-sqlite/

> Unfortunately, I don’t think there’s a way to ALTER a table to make it strict. I think you have to copy the data out of the non-strict table into the strict one.

simonw:
Implementation transcript (GPT-5.6 Sol xhigh in Codex Desktop):  https://gist.github.com/simonw/ab8256b81646ad967a601975e206de64

simonw:
Wrote a bit about this on my blog: https://simonwillison.net/2026/Jul/11/sqlite-utils/

simonw:
Documentation for `.transform()`: https://sqlite-utils.datasette.io/en/stable/python-api.html#changing-strict-mode

PR #788, opened 2026-07-11
.transform(strict=) and sqlite-utils transform --strict/--no-strict

simonw:
Refs:
- #787

----
📚 Documentation preview 📚: https://sqlite-utils--788.org.readthedocs.build/en/788/

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,8 @@
 Unreleased
 ----------
 
+- ``table.transform()`` now accepts ``strict=True`` or ``strict=False`` to change a table's SQLite strict mode. Omitting the option, or passing ``strict=None``, preserves the existing mode. (:issue:`787`)
+- The ``sqlite-utils transform`` command now accepts ``--strict`` and ``--no-strict`` to change a table's SQLite strict mode. Omitting both options preserves the existing mode. (:issue:`787`)
 - ``sqlite-utils query`` can now read the SQL query from standard input by passing ``-`` in place of the query, for example ``echo "select * from dogs" | sqlite-utils query dogs.db -``. (:issue:`765`)
 - ``sqlite-utils insert`` and ``sqlite-utils upsert`` now accept a ``--code`` option for :ref:`providing a block of Python code <cli_insert_code>` (or a path to a ``.py`` file) that defines a ``rows()`` function or ``rows`` iterable of rows to insert, as an alternative to importing from a file. (:issue:`684`)
 - ``sqlite-utils insert`` and ``sqlite-utils upsert`` now accept ``--type column-name type`` to :ref:`override the type automatically chosen when the table is created <cli_insert_csv_tsv_column_types>`. This is useful for CSV or TSV columns such as ZIP codes that look like integers but should be stored as ``TEXT`` to preserve leading zeros. (:issue:`131`)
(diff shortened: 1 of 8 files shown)


simonw:
> use `uv run python -c` and manually exercise the new `.transform(strict=`) option, see if you can find any edge-cases or bugs

GPT-5.6 Sol then found some edge-cases:

> Manual probing found two bugs directly relevant to the new option.
> 
> 1. Explicit `strict=True` silently fails on unsupported SQLite
> 
> Simulating an older SQLite version produced:
> 
> ```python
> t.transform(strict=True)
> t.strict  # False
> ```
> 
> `create_table_sql()` only appends `STRICT` when `db.supports_strict` is true ([db.py](/Users/simon/Dropbox/dev/sqlite-utils/sqlite_utils/db.py:1554)). An explicit conversion should probably raise an informative error instead of succeeding as a no-op.
> 
> 2. The `Table` object retains stale strict defaults
> 
> After:
> 
> ```python
> t = db.table("items", strict=True)
> t.create({"id": int})
> t.transform(strict=False)
> ```
> 
> The schema becomes non-strict, but `t._defaults["strict"]` remains `True`. Dropping and recreating through the same object unexpectedly creates another strict table. The reverse happens when converting from non-strict to strict.
> 
> The default should be updated after a successful transformation, including storing `False`. It should remain unchanged if the transformation rolls back.
> 
> Other notable edge cases:
> 
> - SQLite `ANY` columns are introspected as `float`, so transforming a native strict table containing `ANY` can fail or change it to `REAL`. This is an existing transform/introspection limitation, but applies to `strict=None` and `strict=True`.
> - An incompatible declared default does not prevent conversion. An empty `INTEGER DEFAULT 'not-an-integer'` table becomes strict successfully, but later inserting the default raises `IntegrityError`. This is SQLite behavior.
> - Triggers are dropped during transformation. This is existing documented behavior, not introduced by this change.
> - Losslessly convertible values work: `"123"` and `"123.0"` become integer `123`.
> - Fractional numbers, incompatible text, and BLOBs correctly fail and roll back.
> - Indexes and foreign keys were preserved.
> - Failed conversion inside an existing `db.atomic()` transaction rolled back only the transform; the outer transaction remained usable.
> - `keep_table` correctly retained the original non-strict table alongside the new strict table.
> - Repeated `None → True → None → False → None` transitions behaved correctly.
> 
> No files were changed during this diagnostic pass.

simonw:
I'm going to raise errors if you attempt to convert to STRICT with a SQLite version that fails the `db.supports_strict` test.

simonw:
We don't have any mechanism to support `ANY` columns at the moment. Open question how to deal with that. Options include:

- Ignore the problem entirely
- Add a `sqlite_utils.ANY` constant which can be used in create table calls, e.g. `db.create_table("t", {"id": int, "name": str, "misc": sqlite_utils.ANY})` - would have to be handled in `add_column()` and `transform()` and a bunch of other places too.
- Don't support them in create_table/etc but DO support them in introspection, since that's part of how `transform()` works

simonw:
Here's a reproduction of the problem where the default `strict` value for a table is not correctly updated:
```python

db = Database(memory=True)

table = db.table("items", strict=True)
table.create({"id": int})

table.transform(strict=False)
assert table.strict is False

# Recreate using the same Table object's stale strict=True default:
table.create({"id": int}, replace=True)

assert table.strict is True  # Unexpectedly strict again
```

PR #793, opened 2026-07-12
Fix transform() dropping AUTOINCREMENT from INTEGER PRIMARY KEY columns

### LINKED ISSUE #602: `sqlite-utils transform` removes the `AUTOINCREMENT` keyword
### Context

We ran into this bug randomly, noticing that deleted `ROWID` would get reused after migrating the DB. Using `transform` to change any column in the table will also unexpectedly strip away the `AUTOINCREMENT` keyword from the primary key definition, even if it was not the transformation target.

### Reproducible example

**Original database**

```sql
$ sqlite3 test.db << EOF
CREATE TABLE mytable (
    col1 INTEGER PRIMARY KEY AUTOINCREMENT,
    col2 TEXT NOT NULL
)
EOF

$ sqlite3 test.db ".schema mytable"
CREATE TABLE mytable (
    col1 INTEGER PRIMARY KEY AUTOINCREMENT,
    col2 TEXT NOT NULL
);
```

**Modified database after sqlite-utils**

```sql
$ sqlite-utils transform test.db mytable --rename col2 renamedcol2

$ sqlite3 test.db "SELECT sql FROM sqlite_master WHERE name = 'mytable';"
CREATE TABLE IF NOT EXISTS "mytable" (
   [col1] INTEGER PRIMARY KEY,
   [renamedcol2] TEXT NOT NULL
);
```

ikatyal2110:
Calling .transform() on a table with INTEGER PRIMARY KEY AUTOINCREMENT silently dropped the AUTOINCREMENT keyword, meaning deleted row IDs could be reused in subsequent inserts. This fix detects AUTOINCREMENT in the existing table schema and preserves it in the recreated table by passing it through create_table_sql(). Fixes #602

----
📚 Documentation preview 📚: https://sqlite-utils--793.org.readthedocs.build/en/793/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -1400,6 +1400,7 @@ class Database:
         extracts: Optional[Union[Dict[str, str], List[str]]] = None,
         if_not_exists: bool = False,
         strict: bool = False,
+        autoincrement: bool = False,
     ) -> str:
         """
         Returns the SQL ``CREATE TABLE`` statement for creating the specified table.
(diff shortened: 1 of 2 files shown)


simonw:
Superseded and fixed in 4.2.

PR #809, opened 2026-07-21
Fix rows_from_file() crash on empty or whitespace-only input

### LINKED ISSUE #808: rows_from_file() crashes with csv.Error on empty or whitespace-only files
When `rows_from_file()` is called with an empty file (or a file containing only whitespace) and no explicit `format` argument, it raises an uncaught `csv.Error: Could not determine delimiter`.

**Repro:**
```python
import io
from sqlite_utils.utils import rows_from_file

rows, fmt = rows_from_file(io.BytesIO(b""))
list(rows)  # raises csv.Error: Could not determine delimiter
```

**Root cause:** In `utils.py`, the auto-detect branch strips the first 2048 bytes and passes the result to `csv.Sniffer().sniff()`. When `first_bytes.strip()` is empty (empty file or whitespace only), `sniff("")` raises `csv.Error` rather than returning gracefully.

The fix is to return an empty iterator before reaching `sniff()` when `first_bytes` is empty.

ikatyal2110:
When `rows_from_file()` is called without an explicit format on an empty or whitespace-only file, the auto-detect path calls `csv.Sniffer().sniff("")` which raises `csv.Error: Could not determine delimiter`. The fix adds an early return of an empty iterator when the peeked bytes are empty after stripping, and a regression test covers both the empty-file and whitespace-only cases. Fixes #808

### DIFF
--- a/sqlite_utils/utils.py
+++ b/sqlite_utils/utils.py
@@ -379,6 +379,8 @@ def rows_from_file(
             raise TypeError(
                 "rows_from_file() requires a file-like object that supports peek(), such as io.BytesIO"
             )
+        if not first_bytes:
+            return iter([]), Format.CSV
         if first_bytes.startswith(b"[") or first_bytes.startswith(b"{"):
             # TODO: Detect newline-JSON
             return rows_from_file(buffered, format=Format.JSON)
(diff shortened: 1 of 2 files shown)


simonw:
I landed the more recent PR https://github.com/simonw/sqlite-utils/pull/837

PR #817, opened 2026-07-26
Fix rows_where() crash when offset is given without limit

### LINKED ISSUE #816: rows_where() raises OperationalError when offset is given without limit

**Repro:**
```python
import sqlite_utils

```

ikatyal2110:
SQLite requires a LIMIT clause before OFFSET; without one, `rows_where(offset=N)` generates `SELECT … OFFSET N` which raises `OperationalError: near "N": syntax error`. The fix emits `LIMIT -1` automatically when offset is set but limit is not, which SQLite treats as "no upper bound". The same pattern is fixed in `search_sql()`. A regression test is added. Fixes #816

---
_Generated by [Claude Code](https://claude.ai/code/session_01PZWbR3mvrRvcHsm5PWknZz)_

----
📚 Documentation preview 📚: https://sqlite-utils--817.org.readthedocs.build/en/817/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2003,6 +2003,8 @@ class Queryable:
         if limit is not None:
             sql += f" limit {limit}"
         if offset is not None:
+            if limit is None:
+                sql += " limit -1"
             sql += f" offset {offset}"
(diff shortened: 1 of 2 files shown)


simonw:
I fixed this with:
- #821

PR #819, opened 2026-07-30
fix: allow rows_where(offset=...) without limit

### LINKED ISSUE #816: rows_where() raises OperationalError when offset is given without limit

ltsyk:
## Summary
`rows_where(offset=N)` without `limit` produced invalid SQL because SQLite requires LIMIT with OFFSET.

## Change
When `offset` is set and `limit` is omitted, emit `LIMIT -1` before OFFSET.

Fixes #816

----
📚 Documentation preview 📚: https://sqlite-utils--819.org.readthedocs.build/en/819/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2003,6 +2003,9 @@ class Queryable:
         if limit is not None:
             sql += f" limit {limit}"
         if offset is not None:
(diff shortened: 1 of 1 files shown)


simonw:
I fixed this with:
- #821

PR #821, opened 2026-08-01
Emit LIMIT -1 when offset is used without limit

### LINKED ISSUE #816: rows_where() raises OperationalError when offset is given without limit

ethanhawkes-gif:
Fixes:
- #816.

SQLite requires a `LIMIT` clause to appear before `OFFSET`, so passing `offset` without `limit` generated invalid SQL:

```python
db["t"].rows_where(offset=2)
# select * from "t" offset 2
```

A negative limit means "no upper bound" in SQLite, so emitting `limit -1 offset N` returns all rows from position N onwards.

The issue mentions `rows_where()` and `search_sql()`. There is a third instance of the same pattern in the `sqlite-utils rows` CLI command, which fails the same way:

```
$ sqlite-utils rows data.db t --offset 1
Error: near "1": syntax error
```

So this fixes three sites:

- `Queryable.rows_where()` — also covers `pks_and_rows_where()`, which delegates to it
- `Table.search_sql()` — also covers `search()`
- the `sqlite-utils rows` CLI command

Tests added:

- two new cases in the existing `test_rows_where_offset_limit` parametrize (`offset` with `limit=None`, including `offset=0`)
- `test_pks_and_rows_where_offset_without_limit`
- `test_search_offset_without_limit`
- a new `--offset` without `--limit` case in the `sqlite-utils rows` CLI parametrize

Full suite passes (1379 passed, 16 skipped), plus `black --check`, `flake8` and `mypy` are clean.

----
📚 Documentation preview 📚: https://sqlite-utils--821.org.readthedocs.build/en/821/

### DIFF
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -2478,8 +2478,6 @@ def rows(
     if limit:
         sql += f" limit {limit}"
     if offset:
-        # SQLite requires a limit clause before offset - a negative limit
-        # means "no upper bound", so offset works without an explicit limit
         if not limit:
             sql += " limit -1"
         sql += f" offset {offset}"
--- a/sqlite_utils/cli.py
+++ b/sqlite_utils/cli.py
@@ -2478,6 +2478,10 @@ def rows(
     if limit:
         sql += f" limit {limit}"
     if offset:
+        # SQLite requires a limit clause before offset - a negative limit
+        # means "no upper bound", so offset works without an explicit limit
+        if not limit:
+            sql += " limit -1"
         sql += f" offset {offset}"
     ctx.invoke(
         query,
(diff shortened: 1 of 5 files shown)


ethanhawkes-gif:
Friendly ping on this one — it's mergeable and CI is green, and I'd rather have a clear no than
leave it sitting.

The change is narrow: when `offset` is passed without `limit`, emit `LIMIT -1` so SQLite accepts
the query instead of raising a syntax error. Per the SQLite SELECT docs
(https://sqlite.org/lang_select.html): "If the LIMIT expression evaluates to a negative value,
then there is no upper bound on the number of rows returned." Generated SQL is unchanged whenever
a limit is already present.

Any of these work for me — merge, request changes, or close it as intended behaviour. Just say
which and I'll handle it.

simonw:
Thanks for this!

PR #823, opened 2026-08-02
Fix transform() raising TransformError when renaming an indexed column

### LINKED ISSUE #822: transform(rename=...) raises TransformError when the renamed column has an index
When a table has an index on a column and you rename that column via `transform(rename=...)`, a `TransformError` is raised even though the column still exists under its new name.

```python
import sqlite_utils
db["t"].insert({"id": 1, "name": "Alice"}, pk="id")
db["t"].create_index(["name"])
db["t"].transform(rename={"name": "full_name"})
# TransformError: Index 'idx_t_name' column 'name' is not in updated table 't'.
```

The error says the column "is not in updated table", but a renamed column is by definition still present. The check in `_transform_sql()` conflates `col in rename` (column still exists, just renamed) with `col in drop` (column truly removed). Only dropped columns make an index unrecoverable; renamed columns can have their index recreated using the new name.

ikatyal2110:
When a column that has an index is renamed via `transform(rename=...)`, the index can be recreated using the new column name — the error was wrong to treat renaming the same as dropping. The fix updates the stored index SQL to reference the new column name instead of raising `TransformError`. Fixes #822.

---
_Generated by [Claude Code](https://claude.ai/code/session_01T2B2Q3r6MjX2PQXHQAj4ox)_

----
📚 Documentation preview 📚: https://sqlite-utils--823.org.readthedocs.build/en/823/

### DIFF
--- a/sqlite_utils/db.py
+++ b/sqlite_utils/db.py
@@ -2851,12 +2851,16 @@ class Table(Queryable):
                 if keep_table:
                     sqls.append(f"DROP INDEX IF EXISTS {quote_identifier(index.name)};")
                 for col in index.columns:
-                    if col in rename or col in drop:
+                    if col in drop:
                         raise TransformError(
                             f"Index '{index.name}' column '{col}' is not in updated table '{self.name}'. "
                             f"You must manually drop this index prior to running this transformation "
                             f"and manually recreate the new index after running this transformation. "
                             f"The original index sql statement is: `{index_sql}`. No changes have been applied to this table."
+                        )
+                    elif col in rename:
+                        index_sql = index_sql.replace(
+                            quote_identifier(col), quote_identifier(rename[col])
                         )
                 sqls.append(index_sql)
         return sqls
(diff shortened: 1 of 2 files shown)


sqlite_utils/db.py
```diff
@@ -2851,13 +2851,17 @@ def fk_with_renamed_columns(fk: ForeignKey) -> ForeignKey:
...
                         raise TransformError(
                             f"Index '{index.name}' column '{col}' is not in updated table '{self.name}'. "
                             f"You must manually drop this index prior to running this transformation "
                             f"and manually recreate the new index after running this transformation. "
                             f"The original index sql statement is: `{index_sql}`. No changes have been applied to this table."
                         )
+                    elif col in rename:
+                        index_sql = index_sql.replace(
+                            quote_identifier(col), quote_identifier(rename[col])
+                        )
```
simonw, line 2864: This feels too risk, I'm going to look for a better way.

simonw:
Fixed in https://github.com/simonw/sqlite-utils/commit/57192ef4e36c334bc2946a10547bf64d63621127

Issue #831, opened 2026-08-07
transform() should work for tables that are referenced by views


simonw:
```python
import sqlite_utils

db = sqlite_utils.Database(memory=True)
db["t"].insert({"id": 1, "name": "one"}, pk="id")
db.execute("create view v as select id, name from t")
db["t"].transform(rename={"name": "title"})
# sqlite3.OperationalError: error in view v: no such table: main.t
```

Same for the CLI:

```bash
sqlite-utils transform data.db t --rename name title
```

The fix may look like this:

```sql
CREATE TABLE "t_new_e47fb9e02388" (...);
INSERT INTO "t_new_e47fb9e02388" (...) SELECT ... FROM "t";
DROP TABLE "t";
PRAGMA legacy_alter_table=ON;
ALTER TABLE "t_new_e47fb9e02388" RENAME TO "t";
PRAGMA legacy_alter_table=OFF;
```

<details><summary>Initial suggestion by Fable 5</summary>

Confirmed with sqlite-utils 4.1.1 on SQLite 3.45.1. Every transform-based
operation is affected: retyping, renaming, dropping or reordering columns,
changing primary keys, and all foreign key modifications.

Downstream impact: Datasette's `/db/table/-/alter` API calls `transform()`
directly, so any Datasette table with a dependent view cannot be altered.
datasette-edit-schema worked around this by reading all dependent view
schemas, dropping the views, transforming, then recreating them inside one
transaction — that workaround should live here instead, so every consumer
of `transform()` benefits.

## Root cause

`transform_sql()` generates statements ending in:

```sql
DROP TABLE "t";
ALTER TABLE "t_new_e47fb9e02388" RENAME TO "t";
```

Since SQLite 3.25, `ALTER TABLE ... RENAME TO` validates and rewrites
references to the renamed table in **every** view and trigger definition,
unless `PRAGMA legacy_alter_table` is ON. During that validation step SQLite
tries to resolve view `v`, which references `"t"` — just dropped — and the
whole statement fails with `error in view v: no such table: main.t`.

## Second bug: `keep_table=` silently repoints views at the backup table

When the rename *succeeds*, SQLite's reference-rewriting causes a worse,
silent problem with `keep_table`:

```python
db["t"].transform(types={"name": str}, keep_table="t_backup")
db.execute("select sql from sqlite_master where name = 'v'").fetchone()
# ('CREATE VIEW v as select id from "t_backup"',)
```

[...]

PR #832, opened 2026-08-07
transform() now works for tables referenced by views

simonw:
Closes:
- #831

<details><summary>Claude Fable 5 (Claude Code for web) PR</summary>

## What changed

`transform_sql()` now brackets its `ALTER TABLE ... RENAME TO` statements with `PRAGMA legacy_alter_table=ON`, followed by a closing pragma that restores the value the connection had when the SQL was generated (usually `OFF`, the SQLite default):

```sql
CREATE TABLE "t_new_x123" (...);
INSERT INTO "t_new_x123" (...) SELECT ... FROM "t";
DROP TABLE "t";
PRAGMA legacy_alter_table=ON;
ALTER TABLE "t_new_x123" RENAME TO "t";
PRAGMA legacy_alter_table=OFF;
```

In the `keep_table=` variant the pragma brackets both renames.

## Why

Since SQLite 3.25, `ALTER TABLE ... RENAME TO` validates and rewrites references to the renamed table in every view definition. This caused two bugs:

1. **Transform failed outright** when a view referenced the table being transformed - the rename step raised `error in view v: no such table: main.t`, breaking every transform-based operation (retyping, renaming, dropping columns, changing primary keys, foreign key changes).
2. **`keep_table=` silently repointed views at the backup table** - when the rename succeeded, SQLite rewrote dependent views to select from the frozen backup copy instead of the live table, with no error or warning.

These renames are an internal implementation detail of `transform()`, so view definitions should not be touched. The pragmas are emitted from `transform_sql()` rather than handled inside `transform()`, so the documented "run these statements yourself" workflow and `sqlite-utils transform --sql` output remain standalone-correct.

## Resulting semantics

- Views survive `transform()` with their SQL byte-for-byte unchanged, and keep reading from the live table even with `keep_table=`.
- A view referencing a column the transform renamed or dropped remains defined but raises `no such column` when next queried - inherent to SQLite views, whose SQL is stored as text.
- `PRAGMA legacy_alter_table` is restored to the value it had before the transform, noted in the docs.
- No new parameters, no API changes.

## Reviewer notes

- Tests: seven new tests in `tests/test_transform.py` covering view preservation across rename/retype/pk-change/FK-change transforms, the renamed-column caveat, a view-on-view chain, the `keep_table` repointing regression, standalone execution of `transform_sql()` output, and transform inside an already-open transaction. The 13 parametrized expected-SQL lists gained the two pragma lines. Full suite: 1403 passed, 16 skipped.
- Docs: new "Tables referenced by views" section in `docs/python-api.rst`, cross-reference and updated `--sql` example in `docs/cli.rst`, changelog entry (in an "Unreleased" section - fold into the next release heading when cutting it).
- Also fixed a pre-existing missing blank line in `docs/changelog.rst` (before the `.. _v4_1:` label) that broke strict Sphinx builds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

</details>

----
📚 Documentation preview 📚: https://sqlite-utils--832.org.readthedocs.build/en/832/

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -1986,6 +1986,17 @@ A bare column name drops any foreign key that column participates in, including 
 
 Renaming a column with ``rename=`` updates any foreign keys that use it, and dropping a column with ``drop=`` also drops any foreign keys it participates in - for a compound foreign key this removes the whole constraint.
 
+.. _python_api_transform_views:
+
+Tables referenced by views
+--------------------------
+
+Tables that are referenced by views can be safely transformed - the view definitions are left byte-for-byte unchanged, and views continue to read from the live table even when ``keep_table=`` is used to keep a copy of the original around.
+
+A view that references a column which the transform renamed or dropped will remain defined but will raise a ``no such column`` error when it is next queried. This is inherent to SQLite views, whose SQL is stored as text - if you rename or drop columns that a view depends on you should update that view definition yourself.
+
+To achieve this, the SQL produced by ``transform_sql()`` brackets its ``ALTER TABLE ... RENAME TO`` statements with ``PRAGMA legacy_alter_table=ON`` and ``PRAGMA legacy_alter_table=OFF`` - without this, SQLite would attempt to rewrite references to the renamed table in every view definition, which fails when a view references the table that was just dropped. One consequence is that ``PRAGMA legacy_alter_table`` is reset to ``OFF`` (the SQLite default) after a transform, even if it was previously set to ``ON`` for the connection.
+
 .. _python_api_transform_sql:
 
 Custom transformations with .transform_sql()
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -1995,7 +1995,7 @@ Tables that are referenced by views can be safely transformed - the view definit
 
 A view that references a column which the transform renamed or dropped will remain defined but will raise a ``no such column`` error when it is next queried. This is inherent to SQLite views, whose SQL is stored as text - if you rename or drop columns that a view depends on you should update that view definition yourself.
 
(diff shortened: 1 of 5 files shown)


docs/python-api.rst
```diff
@@ -1986,6 +1986,17 @@ A bare column name drops any foreign key that column participates in, including
...
+.. _python_api_transform_views:
+
+Tables referenced by views
+--------------------------
+
+Tables that are referenced by views can be safely transformed - the view definitions are left byte-for-byte unchanged, and views continue to read from the live table even when ``keep_table=`` is used to keep a copy of the original around.
+
+A view that references a column which the transform renamed or dropped will remain defined but will raise a ``no such column`` error when it is next queried. This is inherent to SQLite views, whose SQL is stored as text - if you rename or drop columns that a view depends on you should update that view definition yourself.
+
+To achieve this, the SQL produced by ``transform_sql()`` brackets its ``ALTER TABLE ... RENAME TO`` statements with ``PRAGMA legacy_alter_table=ON`` and ``PRAGMA legacy_alter_table=OFF`` - without this, SQLite would attempt to rewrite references to the renamed table in every view definition, which fails when a view references the table that was just dropped. One consequence is that ``PRAGMA legacy_alter_table`` is reset to ``OFF`` (the SQLite default) after a transform, even if it was previously set to ``ON`` for the connection.
```
simonw: > One consequence is that `PRAGMA legacy_alter_table is reset to OFF (the SQLite default) after a transform, even if it was previously set to ON` for the connection.

We can do better than that: spot what the previous value is at the start of the transform and set it back to that at the end.

simonw:
Got a test failure against older SQLite 3.23.1 (prior to `PRAGMA legacy_alter_table` being [added in 3.26](https://sqlite.org/releaselog/3_26_0.html)):
```
    def test_transform_restores_legacy_alter_table_setting(fresh_db):
        dogs = fresh_db["dogs"]
        dogs.insert({"id": 1, "name": "Cleo"}, pk="id")
        # Default is OFF, reset to OFF afterwards
        dogs.transform(types={"name": str})
>       assert fresh_db.execute("PRAGMA legacy_alter_table").fetchone()[0] == 0
E       TypeError: 'NoneType' object is not subscriptable
```

PR #840, opened 2026-08-13
Fix count_where() raising OperationalError on nonexistent tables

ikatyal2110:
rows_where() and delete_where() both return gracefully when called on a table that does not yet exist, but count_where() propagates an OperationalError from SQLite instead of returning 0. This adds an existence check at the top of count_where() to make its behaviour consistent with the rest of the API. Adds a test covering count_where() and the count property on a nonexistent table.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WjQWzvQTBBmJeg6czcNbsT)_

----
📚 Documentation preview 📚: https://sqlite-utils--840.org.readthedocs.build/en/840/

### DIFF
--- a/tests/test_introspect.py
+++ b/tests/test_introspect.py
@@ -96,6 +96,12 @@ def test_count_where(existing_db):
     assert existing_db.table("foo").count_where("text != :t", {"t": "two"}) == 2
 
 
+def test_count_where_nonexistent_table(fresh_db):
+    assert fresh_db.table("does_not_exist").count_where() == 0
+    assert fresh_db.table("does_not_exist").count_where("id > ?", [5]) == 0
+    assert fresh_db.table("does_not_exist").count == 0
+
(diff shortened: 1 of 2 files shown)


tests/test_introspect.py
```diff
@@ -96,6 +96,12 @@ def test_count_where(existing_db):
     assert existing_db.table("foo").count_where("text != :t", {"t": "two"}) == 2
 
 
+def test_count_where_nonexistent_table(fresh_db):
+    assert fresh_db.table("does_not_exist").count_where() == 0
+    assert fresh_db.table("does_not_exist").count_where("id > ?", [5]) == 0
```
simonw, line 101: This doesn't feel right to me. I prefer the existing behavior of raising an error.

simonw:
Closed in favor of:
- #841

PR #848, opened 2026-08-26
docs: keep CLI options literal in Python API cross-links

fly1d:
## Summary

Sphinx's smartquotes transform rewrites `--` as an en dash when an option appears inside `:ref:` link text. Issue #493 tracks this rendering problem, and #844 addresses the two affected links in the install section. Six CLI cross-links in `docs/python-api.rst` still had the same issue.

This PR updates only those six notes by keeping the command name as an internal reference and rendering each option as an inline literal. The generated documentation now preserves copyable `--attach`, `--not-null`, `--default`, `--replace`, `--alter`, `--functions`, and `--init-spatialite` flags.

## Verification

- `uv run --group docs sphinx-build -W -b html docs /tmp/sqlite-utils-docs-build`
- `uv run cog --check README.md docs/*.rst`
- `uv run --group docs codespell docs/python-api.rst --ignore-words docs/codespell-ignore-words.txt`
- `git diff --check`
- Rendered HTML inspection confirmed the six links and literal flags.

A full Sphinx linkcheck also ran. It reported two pre-existing unrelated broken links in `docs/contributing.rst` and the existing WGS84 link in `docs/python-api.rst`; neither file location was changed by this PR.

Documentation only; no runtime behavior changes and no changelog entry is needed.

Related to #493. Follow-up to #844.

OpenAI Codex assisted with implementation and verification. Human review remains required.

----
📚 Documentation preview 📚: https://sqlite-utils--848.org.readthedocs.build/en/848/

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -185,7 +185,7 @@ You can attach an additional database using the ``.attach()`` method, providing 
 You can reference tables in the attached database using the alias value you passed to ``db.attach(alias, filepath)`` as a prefix, for example the ``second.table_in_second`` reference in the SQL query above.
 
 .. note::
-    In the CLI: :ref:`sqlite-utils --attach <cli_query_attach>`
+    In the CLI: :ref:`sqlite-utils <cli_query_attach>` ``--attach``
 
 .. _python_api_tracing:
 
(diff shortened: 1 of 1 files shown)


fly1d:
I should have checked with @adrianrfreedman before opening this PR. Their issue comment identified these exact six locations, established the solution, and said they already had the fix locally. Sorry for stepping into work they had offered to submit. I’m closing this so they can open their version if they still want to.

